PR #17, opened 2023-02-20
Update PluginDescriptor to include the`target` field as part of the contract

### LINKED ISSUE #16: Each plugin binary should advertise the `target` information
With the recent changes of introducing the `target` concept in the Tanzu CLI, the uniqueness of the plugin is identified with the `name` and `target` of the plugin.

Currently, the target of the plugin is determined based on the context or the discovery mechanism. However, the plugin itself is not advertising the `target` for which it is built. This makes it challenging to know the plugin's target during build time and to get additional details about the plugins.

Each plugin binary should advertise its `target` information. This can be done as part of the `tanzu <plugin> info` contract by adding an additional required field called `target` in the PluginDescriptor. This will make each plugin develop to provide/configure the `target` field while compiling the plugin using the `builder` plugin.

**Additional context**

anujc25:

This change updates the plugin contract to specify `target` information as part of the plugin descriptor. Old plugins will continue to work but new plugins using the latest plugin runtime will be enforced to provide target information as part of PluginDescriptor.

This change doesn't affect existing plugins because CLI is consuming this `target` field during the build time of the plugins and not during the runtime to determine the `target` of the plugin. So, once the plugins are already built, they should continue to work with the newer version of `tanzu-cli`.

Renames `TargetNone` to `TargetUnknown` and added another target `TargetGlobal`.

While installing the plugin this change will be made backward compatible.

Fixes #16

- Updated unit tests.
- Verified that the build fails if a user is not providing the `Target` of the plugin as part of the PluginDescriptor.
- Tested the changes along with Tanzu Core CLI

### Release note

```release-note
Update PluginDescriptor to include the `target` field as a mandatory field as part of the contract
```

### DIFF
--- a/config/types/clientconfig.go
+++ b/config/types/clientconfig.go
@@ -15,15 +15,20 @@ type Target string
 
 const (
 	// TargetK8s is a kubernetes target of the CLI
+	// This target applies if the plugin is interacting with a Kubernetes cluster
 	TargetK8s Target = "kubernetes"
 	targetK8s Target = "k8s"
 
 	// TargetTMC is a Tanzu Mission Control target of the CLI
+	// This target applies if the plugin is interacting with a Tanzu Mission Control endpoint
 	TargetTMC Target = "mission-control"
 	targetTMC Target = "tmc"
 
-	// TargetNone is used for plugins that are not associated with any target
-	TargetNone Target = ""
+	// TargetGlobal is used for plugins that are not associated with any target
+	TargetGlobal Target = "global"
+
+	// TargetUnknown specifies that the target is not currently known
+	TargetUnknown Target = ""
 )
 
 var (
--- a/config/types/clientconfig_helper.go
+++ b/config/types/clientconfig_helper.go
@@ -9,16 +9,22 @@ func StringToTarget(target string) Target {
 		return TargetK8s
 	} else if target == string(targetTMC) || target == string(TargetTMC) {
 		return TargetTMC
-	} else if target == string(TargetNone) {
-		return TargetNone
+	} else if target == string(TargetGlobal) {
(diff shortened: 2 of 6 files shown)


marckhouzam (approved):
Very nice.  Thanks!
LGTM

vuil:
> Old plugins will continue to work 
@anujc25 can you update in the PR description with some explanation of what this effectively means (on why it doesn't matter for older plugins, regardless of whether they were meant to a specific target or not)

anujc25:
> @anujc25 can you update in the PR description with some explanation of what this effectively means (on why it doesn't matter for older plugins, regardless of whether they were meant to a specific target or not)

@vuil I have updated the PR description to specify why it doesn't matter for old plugins.

config/types/clientconfig.go
```diff
@@ -22,8 +22,11 @@ const (
 	TargetTMC Target = "mission-control"
 	targetTMC Target = "tmc"
 
-	// TargetNone is used for plugins that are not associated with any target
-	TargetNone Target = ""
+	// TargetGlobal is used for plugins that are not associated with any target
```
vuil, line 25: Nit: Given that some folks might be new to the target concept and now having to address it in their plugin impl for the first time, besides adding dev md docs _soon_, it would be good when they likely get here to find out what target values to use to have a bit more explanation to each. 

As far as specific targets there are currently only two, kubernetes and mission-control with another two being respective aliases. 
The comment should be clear on when kubernetes should be used (something like "when the plugin should only apply when interacting with a Kubernetes cluster"...), when mission-control should be used,

vuil (approved):
lgtm, some suggestions about enhancing a few comments.

PR #24, opened 2023-03-01
Implement `logger`

anujc25:

- Implement support for `logger` 
- This logger can be consumed by plugins to have consistent logging experience across tanzu-cli and plugins
- This has support for level-based logging
- It doesn't support colored logging at the moment
- This `logger` implementation is based on https://github.com/vmware-tanzu/tanzu-framework/tree/main/tkg/log

Fixes #

### Release note

```release-note
Add support for logger which can be consumed by the plugins for the logging to have a consistent logging experience
```


marckhouzam (commented):
One comment about `Errorf` inline.

Out of curiosity, why did you remove `SetChannel()` and `SendProgressUpdate()` which were in the logger in TF?

chandrareddyp (approved):
LGTM

anujc25:
> Out of curiosity, why did you remove SetChannel() and SendProgressUpdate() which were in the logger in TF?

That section was more for specific use-case of sending logs to the kickstart UI. So, I felt it doesn't make sense to add it to this generic logger.

log/log.go
```diff
@@ -0,0 +1,164 @@
...
+	l.Print(msg, nil, "INFO")
+}
+
+// Error logs an error message with the given key/value pairs as context.
+func Error(err error, msg string, kvs ...interface{}) {
+	l.Print(msg, err, "ERROR", kvs...)
+}
+
+// Errorf logs a error message with the given key/value pairs as context.
+func Errorf(format string, args ...interface{}) {
```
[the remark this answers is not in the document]
anujc25: Thanks. Updated.

log/writer.go
```diff
@@ -0,0 +1,196 @@
...
+
+	// write to logfile only if verbosityLevel is <= default VerbosityLevel
+	if logVerbosity <= defaultVerbosity {
+		if w.logFile != "" {
+			fileWriter(w.logFile, fullMsg)
+		}
+	}
+
+	// If showHeader is set to true, log fullMsg with header
+	// If log type if OUTPUT skip the header
```
marckhouzam, line 119:
```suggestion
	// If log type is OUTPUT skip the header
```

marckhouzam (changes requested):
One suggested fix and one thing to possibly discuss

log/writer.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+var (
+	originalStdout *os.File
+	originalStderr *os.File
+)
+
+var defaultVerbosity int32 = 4
+
+// NewWriter returns new custom writter for tkg-cli
```
marckhouzam, line 51:
```suggestion
// NewWriter returns new custom writer
```

log/logger.go
```diff
@@ -0,0 +1,311 @@
...
+		path := file
+		file = path[slash+1:]
+	}
+
+	now := time.Now()
+	_, month, day := now.Date()
+	hour, minute, second := now.Clock()
+
+	// mmdd hh:mm:ss.uuuuuu file:line]
+	return fmt.Sprintf("%02d%02d %02d:%02d:%02d.%06d %s:%d] ", int(month), day, hour, minute, second, now.Nanosecond()/nanoSecondsPerMicroSecond, file, line)
```
marckhouzam, line 310: There is a closing brace but no opening one in this header.

log/logger.go
```diff
@@ -0,0 +1,315 @@
...
+}
+
+// logger defines a clusterctl friendly logr.Logger
+type logger struct {
+	threshold  *int32
+	level      int32
+	prefix     string
+	values     []interface{}
+	callDepth  int
+	showCaller bool
```
marckhouzam, line 45: There is currently no API to set this new variable.
But it is ok, we've spent enough time on this PR.  We can add it in another PR.

log/logger.go
```diff
@@ -0,0 +1,315 @@
...
+		file = "???"
+		line = 1
+	} else if slash := strings.LastIndex(file, "/"); slash >= 0 {
+		path := file
+		file = path[slash+1:]
+	}
+	rfct := time.Now().Format(time.RFC3339)
+	header := fmt.Sprintf("%s ", rfct)
+	if l.showCaller {
+		header = fmt.Sprintf("%s%s:%d] ", header, file, line)
```
marckhouzam, line 312:
```suggestion
		header = fmt.Sprintf("%s[%s:%d] ", header, file, line)
```

marckhouzam (approved):
Thanks for addressing all the comments @anujc25.
Very nice improvements you've made!

log/logger.go
```diff
@@ -0,0 +1,315 @@
...
+type logger struct {
+	threshold  *int32
+	level      int32
+	prefix     string
+	values     []interface{}
+	callDepth  int
+	showCaller bool
+}
+
+var _ LoggerImpl = &logger{}
```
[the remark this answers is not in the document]
anujc25: It is used to verify `logger` implements all the `LoggerImpl` interface methods.

In this case, we are doing `NewLogger()` so, that function will throw an error if the `logger` is not implementing any `LoggerImpl` function. However, sometimes, you write an interface implementation that should be consumed externally and not internally, it is a good practice to write this statement to verify the implementation is not missing any method implementation required by the interface.

log/logger.go
```diff
@@ -0,0 +1,315 @@
...
+func (l *logger) getLogTypeIndicator(logType string) string {
+	switch logType {
+	case logTypeINFO:
+		return "[i] "
+	case logTypeWARN:
+		return "[!] "
+	case logTypeERROR:
+		return "[x] "
+	case logTypeSUCCESS:
+		return "[✔] "
```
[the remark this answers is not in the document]
anujc25: Replaced with `[ok]`.

log/logger.go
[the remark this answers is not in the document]
marckhouzam: This is a great trick!  I wasn't aware of it.  Thanks!

PR #34, opened 2023-03-21
Add Deprecation annotation to the Deprecated APIs

anujc25:

- This PR marks all the server APIs for the configuration file as deprecated and suggests using context APIs instead.
- This PR also marks additional ClientConfig configurations that are deprecated and is not used by the core-cli as Deprecated

- The PR uses the standard deprecation syntax for the APIs as specified in https://go.dev/blog/godoc

> To signal that an identifier should not be used, add a paragraph to its doc comment that begins with "Deprecated:" followed by some information about the deprecation.

- The documentation site [pkg.go.dev](https://pkg.go.dev/) hides the documentation for deprecated identifiers behind a click of a "show" button.

- The [staticcheck](https://staticcheck.io/) tool reports use of deprecated identifiers

Fixes #

### Release note

```release-note
Add Deprecation annotation to the Deprecated APIs
```

### DIFF
--- a/config/servers.go
+++ b/config/servers.go
@@ -14,6 +14,8 @@ import (
 )
 
 // GetServer retrieves server by name
+//
+// Deprecated: This API is deprecated. Use GetContext instead.
 func GetServer(name string) (*configtypes.Server, error) {
 	// Retrieve client config node
 	node, err := getClientConfigNode()
@@ -24,12 +26,16 @@ func GetServer(name string) (*configtypes.Server, error) {
 }
 
 // ServerExists checks if server by specified name is present in config
+//
+// Deprecated: This API is deprecated. Use ContextExists instead.
 func ServerExists(name string) (bool, error) {
 	exists, _ := GetServer(name)
 	return exists != nil, nil
 }
 
 // GetCurrentServer retrieves the current server
+//
+// Deprecated: This API is deprecated. Use GetCurrentContext instead.
 func GetCurrentServer() (*configtypes.Server, error) {
 	// Retrieve client config node
 	node, err := getClientConfigNode()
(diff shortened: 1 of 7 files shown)


mpanchajanya (commented):
There are also few server related methods in clientconfig.go  that we could mark deprecated in the similar way.

config/servers.go
```diff
@@ -108,16 +118,22 @@ func RemoveCurrentServer(name string) error {
...
 // PutServer add or update server and currentServer
+//
+// Deprecated: This API is deprecated. Use SetContext instead.
 func PutServer(s *configtypes.Server, setCurrent bool) error {
 	return SetServer(s, setCurrent)
 }
 
 // AddServer add or update server and currentServer
+//
+// Deprecated: This API is deprecated. Use SetContext instead.
```
marckhouzam, line 129: We have a `AddContext` API which may be more intuitive than `SetContext` here, or maybe put both?

marckhouzam:
> There are also few server related methods in clientconfig.go that we could mark deprecated in the similar way.

Agreed.  And looking at those, I noticed there are some types also in `clientconfig_types.go` that are marked for deprecation but they are not using the proper GoDoc format.  Could we change those too?

marckhouzam (approved):
This LGTM but maybe first update the other "Deprecation targeted..." strings to follow proper GoDoc?

config/servers.go
anujc25: Good point. Updated

anujc25:
> There are also few server related methods in clientconfig.go that we could mark deprecated in the similar way.

Thanks for catching this. Marked additional stuff in `clientconfig.go` and `clientconfig_types.go` as deprecated. cc @marckhouzam @mpanchajanya

mpanchajanya (approved):

marckhouzam (approved):
> > There are also few server related methods in clientconfig.go that we could mark deprecated in the similar way.
> 
> Thanks for catching this. Marked additional stuff in `clientconfig.go` and `clientconfig_types.go` as deprecated. cc @marckhouzam @mpanchajanya

Thanks for adding the other ones as well.  It will pave the way to getting rid of them 👏 

LGTM

PR #36, opened 2023-03-26
Disable autogenerated footer in generated docs

vuil:
Small change to disable the autogenerate footer at the end of output from `generate-docs`

Related to the changes made in https://github.com/vmware-tanzu/tanzu-cli/pull/135 so generated docs are more consistent across plugins and core commands and are not subjected to date stamp diffs.

Fixes #N/A

Built plugins with this change. Verified that <plugin-binary> `generate-docs` produces markdown in docs/cli/commands that does not contain above-mentioned footer.

### Release note

```release-note
Markdown generated by `generate-docs` command no long contains `Autogenerated...` footer
```

### DIFF
--- a/plugin/root.go
+++ b/plugin/root.go
@@ -33,5 +33,8 @@ func newRootCmd(descriptor *PluginDescriptor) *cobra.Command {
 		newInfoCmd(descriptor),
 	)
 
+	// Disable footers in docs generated
+	cmd.DisableAutoGenTag = true
+
 	return cmd
 }


plugin/root.go
```diff
@@ -33,5 +33,8 @@ func newRootCmd(descriptor *PluginDescriptor) *cobra.Command {
 		newInfoCmd(descriptor),
 	)
 
+	// Disable footers in docs generated
+	cmd.DisableAutoGenTag = true
```
marckhouzam, line 37: Nit: why not add it in the structure above like the other fields?

marckhouzam (approved):
LGTM
Optional nit.

plugin/root.go
vuil: Makes sense. Updated.

marckhouzam:
actually, when are we going to generate such docs for plug-ins? With the central repository containing a whole bunch of plug-ins are we going to generate the docs for all of them? Meaning, do we need to install all plugins in one machine at one time? and then how do we update those docs? 

I'm not clear about the use case when itt comes to plugins...

PR #47, opened 2023-04-06
Fix Test plugin to include Target by default

### LINKED ISSUE #48: Test plugins created with new plugin-runtime are failing to execute
**Bug description**

Created a new helloworld plugin bootstrapped wth 

`tanzu builder cli add-plugin <plugin-name>` command  and was able to successfully build and install the plugin locally.  However when I tried to run the test plugin it fails

```
❯ tanzu test plugin helloworldplugin
[x] : invalid PluginDescriptor specified: plugin "helloworldplugin-test": target is not valid
Error: exit status 1
Usage:
tanzu test plugin helloworldplugin [flags]
```

**Expected behavior**

Should be able to run the test plugin.

**Steps to reproduce the bug / Relevant debug output**

**Plugin Runtime version (also obtainable from `tanzu <pluginName> info`)**
v0.81.0

**Environment where the bug was observed (cloud, OS, etc)**

anujc25:

- Fix the Test plugin to include Target by default

Without the fix, when we execute the test plugin binary with `tanzu test plugin hello-world` it throws the below error: `invalid PluginDescriptor specified: plugin "hello-world-test": target is not valid`

Fixes #48

### Release note

```release-note
Fix the Test plugin to include Target by default
```

### DIFF
--- a/test/framework/framework.go
+++ b/test/framework/framework.go
@@ -20,6 +20,7 @@ import (
 	"github.com/spf13/pflag"
 	"gopkg.in/yaml.v3"
 
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
 	"github.com/vmware-tanzu/tanzu-plugin-runtime/log"
 	"github.com/vmware-tanzu/tanzu-plugin-runtime/plugin"
 )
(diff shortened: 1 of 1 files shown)


prkalle (approved):
LGTM. Thanks!

vuil (approved):
Thanks for the quick fix, lgtm.

PR #55, opened 2023-04-17
Use generics to convert objects to config nodes

marckhouzam:

This does some code cleanup to take advantage of generics to avoid code duplication.
It replaces the 6 different `convert*ToNode(..)` functions to using a single one with generics: `convertObjectToNode[T](..)`.

Fixes #

Built the CLI using this change and did a `tz config get` after creating contexts.

### Release note

```release-note

```


mpanchajanya (approved):
lgtm! Thanks for addressing this

PR #56, opened 2023-04-18
Update golangci-lint to 1.52.2

marckhouzam:
⚠️  you must remove the old version of the linter to properly get this change locally: `rm -rf hack/tools/bin/golangci-lint`

I wanted to use generics in #55 but the golangci-lint version we are using is too old and gocritic panics.
This commit updates the version to 1.52.2.

Note that "allow-leading-space" is no longer supported as mentioned in https://github.com/golangci/golangci-lint/issues/3063#issuecomment-1206178658 and therefore has been removed.  All instances of `// nolint` have been update to remove the space.

A couple of new warnings were also fixed.

Finally, I have rebase #55 and confirmed we can use generics without linter warnings when using this PR.

Fixes # N/A

make go-lint

I also tried #55 which is introducing generics and ran `make go-lint`.  Before this PR, the linter panics  but with this PR it all passes.

### Release note

```release-note

```


mpanchajanya (approved):
LGTM. Thanks for doing this

anujc25 (approved):

PR #57, opened 2023-04-24
Bug Fix:  Add empty checks for config APIs

mpanchajanya:
-  Without these empty checks the config items like contexts, servers etc duplicate in config files when name is empty
- Added empty value checks for config APIs to return error when encounter an empty value.

Fixes #

Added Unit tests
Updated the Integration tests
Cross-version compatibility tests passed

### Release note

```release-note
added empty checks for config apis. Config APIs now doesn't allow bad data to persist in the config yaml files
```


anujc25 (approved):
LGTM. Thanks!

PR #62, opened 2023-04-25
Add configuration APIs for Certificate configuration

prkalle:
This PR is to add configuration APIs for custom certificate configuration

The sample configuration is shown below which would be added to the configuration file (config-ng.yaml)

```
certs:
    - hostName: localhost:9876
      skipCertVerify: "true"
      insecure: "true"
      caCertData: <REDACTED>
```

Fixes #

Unit tests and CI passed

### Release note

```release-note
Add configuration APIs for Certificate configuration
```

### DIFF
--- a/config/certs.go
+++ b/config/certs.go
@@ -1,0 +1,200 @@
+// Copyright 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package config
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
+
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/config/nodeutils"
+
+	configtypes "github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
+)
+
+// GetCerts retrieves all the certs
+func GetCerts() ([]*configtypes.Cert, error) {
+	// Retrieve client config node
+	node, err := getClientConfigNode()
+	if err != nil {
+		return nil, err
+	}
+
+	return getCerts(node)
+}
+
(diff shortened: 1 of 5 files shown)


config/certs.go
```diff
@@ -0,0 +1,181 @@
...
+		return nil, err
+	}
+	if cfg.Certs != nil {
+		return cfg.Certs, nil
+	}
+	return nil, errors.New("certs not found")
+}
+
+func getCert(node *yaml.Node, hostName string) (*configtypes.Cert, error) {
+	cfg, err := convertNodeToClientConfig(node)
```
mpanchajanya, line 99: nit; pls add hostName validation steps in Get/Delete/Set methods especially if its value is "". so we don't end up have invalid certs like certs without hostname

config/certs_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+func TestSetCert(t *testing.T) {
+	// Setup config data
+	_, cleanUp := setupTestConfig(t, &CfgTestData{})
+
+	defer func() {
+		cleanUp()
+	}()
+
+	tests := []struct {
```
mpanchajanya, line 72: pls add few negative test cases where Set/Get/Delete return errors
when invalid cert data is used to set/delete/get

mpanchajanya (changes requested):
Changes LGTM!. nit! on validating the input args and negative test cases

config/certs.go
prkalle: Good point. Updated. Thanks

config/certs_test.go
prkalle: Updated.

mpanchajanya (approved):

PR #64, opened 2023-05-01
Update Certificate configuration types to use `Host` instead of `HostName`

prkalle:
This PR updates the Certficate configuration types to use `Host` instead of `HostName`.
This is to support both `hostname` and `hostname:port` combinations

Fixes #
Unit tests and CI passed

### Release note

```release-note
N/A
```


anujc25 (approved):
LGTM

PR #76, opened 2023-05-24
 Provide experimental SyncPluginsForTarget API

### LINKED ISSUE #75: Provide API to trigger syncing of plugins for new or updated Context

As a plugin developer, I would like my plugin command implementation that creates or updates a Context to ensure that the plugins that are discoverable via said context are automatically installed on the CLI when said context becomes an active one.

Without an API to ensure the above, the plugin would have to resort to explicitly invoke another CLI command directly (e.g. 'tanzu plugin sync') to achieve the same effect, or implore that the user performs the equivalent

**Additional context**

vuil:

   
    SyncPluginsForTarget will attempt to install plugins required by the active
    Context of the provided target. This is most useful for any plugin
    implementation which creates a new Context or updates an existing one as
    part of its operation, and prefers that the plugins appropriate for the
    Context are immediately available for use.

    Note: This API is considered EXPERIMENTAL. Both the function's signature and
    implementation are subjected to change/removal if an alternative means to
    provide equivalent functionality can be introduced.

---

In addition: 

    Check for alternative command invocation

    Invoke what would be a hidden '_custom_command' command to obtain an
    alternate means to perform, in this case, the plugin syncing operation.
    This command is expected to fail at the moment, and the behavior in
    this situation would be to fall back to the original command.

Fixes #75

- `make test`  (which is updated with test cases to simulate invoking plugin sync or the hypothetical alternate command being discovered and in turn invoked).
- built a plugin linking to this runtime implementation and verified that a plugin command calling SyncPluginsForTarget will result in `tanzu plugin sync` being performed. 
```
tanzu myplugin fakecreate foo
Fake create cluster foo
Sync-ing plugins...
SYNC OK: output=
[i] Checking for required plugins...
[i] Installing plugin 'cluster:v0.28.0' with target 'kubernetes'
[i] Installing plugin 'feature:v0.28.0' with target 'kubernetes'
[i] Installing plugin 'kubernetes-release:v0.28.0' with target 'kubernetes'
[i] Successfully installed all required plugins
[ok] Done
```

### Release note

```release-note
 Provide experimental SyncPluginsForTarget API
```

### DIFF
--- a/plugin/sync_plugins.go
+++ b/plugin/sync_plugins.go
@@ -1,0 +1,44 @@
+// Copyright 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package plugin
+
+import (
+	"bytes"
+	"os"
+	"os/exec"
+
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
+)
+
+// SyncPluginsForTarget will attempt to install plugins required by the active
+// Context of the provided target. This is most useful for any plugin
+// implementation which creates a new Context or updates an existing one as
+// part of its operation, and prefers that the plugins appropriate for the
+// Context are immediately available for use.
+//
+// Note: This API is considered EXPERIMENTAL. Both the function's signature and
+// implementation are subjected to change/removal if an alternative means to
+// provide equivalent functionality can be introduced.
+//
+// Stdout output of the command is returned as a string on successful
+// execution of the command, otherwise, the Stderr output is returned instead.
+func SyncPluginsForTarget(target types.Target) (string, error) {
+	// For now, the implementation expects env var TANZU_BIN to be set and
+	// pointing to the core CLI binary used to invoke the plugin sync with.
+
+	var stderr bytes.Buffer
+	var stdout bytes.Buffer
+
+	cliPath := os.Getenv("TANZU_BIN")
+	command := exec.Command(cliPath, "plugin", "sync")
+	command.Stdout = &stdout
+	command.Stderr = &stderr
+
+	err := command.Run()
+
+	if err != nil {
+		return stderr.String(), err
+	}
+	return stdout.String(), nil
+}
--- a/plugin/sync_plugins.go
+++ b/plugin/sync_plugins.go
@@ -5,11 +5,33 @@ package plugin
 
 import (
 	"bytes"
+	"errors"
+	"fmt"
 	"os"
 	"os/exec"
+	"strings"
 
 	"github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
 )
+
+const (
+	// customCommandName is the name of the command expected to be implemented
+	// by the CLI should there be a need to discover and alternative invocation
+	// method
+	customCommandName string = "_custom_command"
+)
+
+func runCommand(commandPath string, args []string) (bytes.Buffer, bytes.Buffer, error) {
+	var stderr bytes.Buffer
+	var stdout bytes.Buffer
+
+	command := exec.Command(commandPath, args...)
+	command.Stdout = &stdout
+	command.Stderr = &stderr
+
+	err := command.Run()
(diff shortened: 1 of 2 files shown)


plugin/sync_plugins.go
```diff
@@ -0,0 +1,75 @@
...
+	command := exec.Command(commandPath, args...)
+	command.Stdout = &stdout
+	command.Stderr = &stderr
+
+	err := command.Run()
+
+	if err != nil {
+		return stderr, err
+	}
+	return stdout, nil
```
marckhouzam, line 36: So our logs are printed on stderr, so the output of a successful "plugin sync" will only have printouts on stderr.
If the plugin calling this new API wants to parse the list of plugins that was installed, they will not be able to because we won't be returning stderr.

Why not always return stderr and stdout?  That way, you're covered for the future.

vuil: Thanks, nice catch. For this helper function, I agree with the suggestion. But do we extend that reasoning to the Sync API? I am still thinking if it is better to keep the API simple, because it is not entirely obvious that for a successful call that the useful output is in stderr. Perhaps in this particular case, we can retain the same signature, but return the merged result of stdout/stderr instead. Thoughts?

plugin/sync_plugins.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// Check if there is an alternate means to perform the plugin syncing
+	// operation, if not fall back to `plugin sync`
+	output, err := runCommand(cliPath, altCommandArgs)
+	if err == nil {
+		args = strings.Fields(output.String())
+	}
+
+	// Runs the actual command
+	return runCommand(cliPath, args)
```
anujc25, line 74: Does this mean we always invoke 2 tanzu-cli commands even if the `_custom_command` is not available or not implemented?

plugin/sync_plugins.go
marckhouzam: Sounds good to me. 👍

vuil: Cool. I updated the helper function. Also tweak the signature of SyncPluginsForTarget to return a string output instead.

plugin/sync_plugins.go
anujc25: I think that is fine. I just wanted to understand it better to make sure. Thanks for confirming.

plugin/sync_plugins.go
```diff
@@ -0,0 +1,72 @@
...
+	altCommandArgs := []string{customCommandName}
+	args := []string{"plugin", "sync"}
+
+	altCommandArgs = append(altCommandArgs, args...)
+	altCommandArgs = append(altCommandArgs, "--target", string(target))
+
+	// Check if there is an alternate means to perform the plugin syncing
+	// operation, if not fall back to `plugin sync`
+	output, _, err := runCommand(cliPath, altCommandArgs)
+	if err == nil {
```
anujc25, line 65:
```suggestion
	if err == nil && output != "" {
```

plugin/sync_plugins.go
```diff
@@ -0,0 +1,72 @@
...
+		args = strings.Fields(output.String())
+	}
+
+	// Runs the actual command
+	stdoutOutput, stderrOutput, err := runCommand(cliPath, args)
+	return fmt.Sprintf("%s%s", stdoutOutput.String(), stderrOutput.String()), err
```
anujc25, line 71: So, let's say the command writes the logs as mentioned below (note stdout and stderr are just for reference to explain where the log is getting printed)

```
stdout: log1
stderr: log2
stdout: log3
stderr: log4
```

Based on the current implementation, we will be returning something like this as part of this function which changes the log sequence.
```
log1
log3
log2
log4
```

This doesn't seem equivalent to `CombinedOutput` AFAIK. @vuil @marckhouzam please correct me if I am wrong here.

anujc25 (approved):
LGTM

PR #80, opened 2023-07-12
Add CSPOrgID and EntitlementAccountNumber to telemetry options in the config 

prkalle:
This PR adds `cspOrgID` and `entitlementAccountNumber` to the telemetry options in the config

Fixes #

Unit tests and CI checks passed

### Release note

```release-note
Add CSPOrgID and EntitlementAccountNumber to telemetry options in the config 
```

### DIFF
--- a/config/cli_options_telemetry_test.go
+++ b/config/cli_options_telemetry_test.go
@@ -20,11 +20,19 @@ func TestSetGetTelemetryOptions(t *testing.T) {
 	}()
 
 	telOptions := &configtypes.TelemetryOptions{
-		Source: "/fake/path",
+		Source:                   "/fake/path",
+		CSPOrgID:                 "fake-csp-org-id",
+		EntitlementAccountNumber: "fake-entitlement-account-number",
 	}
 
 	telOptionsUpdate := &configtypes.TelemetryOptions{
 		Source: "/fake/path/updated",
+	}
+
+	telOptionsUpdate2 := &configtypes.TelemetryOptions{
+		Source:                   "/fake/path/updated-2",
+		CSPOrgID:                 "fake-csp-org-id-2",
+		EntitlementAccountNumber: "fake-entitlement-account-number-2",
 	}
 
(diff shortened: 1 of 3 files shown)


docs/config.md
```diff
@@ -83,6 +83,13 @@ func GetEnvConfigurations() map[string]string
 func GetEdition() (string, error)
 func SetEdition(val string) (err error)
 
+// Telemetry APIs
```
prkalle, line 86: I hope adding the telemetry APIs in the docs is fine, though these current APIs are used by CLI and telemetry plugin only.

anujc25 (approved):
LGTM

config/cli_options_telemetry_test.go
```diff
@@ -20,13 +20,21 @@ func TestSetGetTelemetryOptions(t *testing.T) {
...
+		EntitlementAccountNumber: "fake-entitlement-account-number",
 	}
 
 	telOptionsUpdate := &configtypes.TelemetryOptions{
 		Source: "/fake/path/updated",
 	}
 
+	telOptionsUpdate2 := &configtypes.TelemetryOptions{
+		Source:                   "/fake/path/updated-2",
+		CSPOrgID:                 "fake-csp-org-id-2",
```
mpanchajanya, line 34: fyi. if empty "" fields are passed to update. APIs will not update the config if value is already present.  We need to use replace patch strategy on those specific fields

mpanchajanya (approved):
lgtm! fyi. if empty "" fields are passed to update. APIs will not update the config if value is already present. We need to use replace patch strategy on those specific fields

prkalle:
> lgtm! fyi. if empty "" fields are passed to update. APIs will not update the config if value is already present. We need to use replace patch strategy on those specific fields

Thanks! I think users should do a GET, followed by DELETE and then SET( the specific field to empty and rest unchanged) inorder to set the filed to empty(delete).

marckhouzam (approved):
LGTM

PR #81, opened 2023-07-21
Introduce additional metadata field to Context

mpanchajanya:
 - At the moment, Tanzu Plugin Runtime - Context configuration does not provide a means to add any additional metadata.
This proposal introduces a new AdditionalMetadata map field to Context to be able to provide additional data for each context. This way we don't have to release a new version of runtime every time we add a new property to the context object.
Example: TMC will need to store a few additional properties for each context and the fields could change in future. In TMC Self-Managed with VCD's login flow, it requires "VCD org" and "VCD OIDC client ID" to be persisted in context, so that TMC Plugins can auto-refresh or re-trigger the token flow if needed when a TMC plugin command is called.

Example usage:
- Add two new properties to context

```
ctx := &Context {
 Name: tanzu-ctx,
 AdditionalMetadata : map[string]interface{
  “token” : “test-token”
  “issuer” : “test-issuer”
 }
}
SetContext(ctx)  // ctx will  now have both token and issuer metadata

```

- Update the existing additional metadata of a context

```
# To merge the additional metadata

ctx := GetContext()
ctx.AdditionalMetadata[“token”]=”updated-token”
SetContext(ctx)  // ctx will now have both issuer and token metadata


# To replace the additional metadata

ctx := GetContext()
ctx.AdditionalMetadata := map[string]interface{}{
 “token”=”only-token”,
}
SetContext(ctx) // Ctx will only have token metadata

# To delete the entire additional metadata

ctx := GetContext()
ctx.AdditionalMetadata := map[string]interface{}{}
SetContext(ctx) // Ctx will not have any additional metadata

```

Fixes #

Added unit tests and integrated tests

### Release note

```release-note
 Introduce new additionalMetadata field in Context
```

### DIFF
--- a/config/config_keys.go
+++ b/config/config_keys.go
@@ -28,4 +28,5 @@ const (
 	KeyTelemetry               = "telemetry"
 	KeyCLIId                   = "cliId"
 	KeySource                  = "source"
+	KeyAdditionalMetadata      = "additionalMetadata"
 )
--- a/config/contexts.go
+++ b/config/contexts.go
@@ -14,33 +14,6 @@ import (
 	configtypes "github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
 )
 
-// ContextOpts is used to construct options for context apis
-type ContextOpts struct {
-	DeleteAllAdditionalMetadata bool   // To delete the entire additional metadata property of context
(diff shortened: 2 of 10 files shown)


anujc25 (approved):
LGTM. Thanks

PR #83, opened 2023-07-28
Expose `TANZU_CLI_LOG_LEVEL` environment variable to control the log level

anujc25:

Tanzu Plugin Runtime exposes a logger functionality and also allows logs to be displayed based on the different log levels.
It also exposes API to control the log level for external plugins.

However, we wanted the log levels to be controlled by the user at the runtime so that we can enable more logs for debugging purposes. This PR provides an environment variable `TANZU_CLI_LOG_LEVEL` to control the log level.

The default value of the Tanzu CLI log level is `3`. So any user wanting to see more debug logs should configure `TANZU_CLI_LOG_LEVEL` with value `>3`. 

Fixes #

* Manually tested the change.
* Added unit tests.
* Sample tests details mentioned under: https://github.com/vmware-tanzu/tanzu-cli/pull/437

### Release note

```release-note
Expose the `TANZU_CLI_LOG_LEVEL` environment variable to control the log level
```

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -8,10 +8,15 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/pkg/errors"
+)
+
+const (
+	EnvTanzuCLILogLevel = "TANZU_CLI_LOG_LEVEL"
 )
 
 var defaultLogThreshold int32 = 3
@@ -28,14 +33,27 @@ type logEntry struct {
 	Values []interface{}
 }
 
-// NewLogger returns a new instance of the clusterctl.
+// NewLogger returns a new instance of the logger.
 func NewLogger() LoggerImpl {
+	logThreshold := getLogThreshold()
 	return &logger{
-		threshold: &defaultLogThreshold,
+		threshold: &logThreshold,
 	}
 }
 
-// logger defines a clusterctl friendly logr.Logger
+func getLogThreshold() int32 {
+	reqLogLevelStr := os.Getenv(EnvTanzuCLILogLevel)
+	if reqLogLevelStr != "" {
+		requestedLogLevel, err := strconv.ParseUint(reqLogLevelStr, 10, 32)
+		if err == nil {
+			return int32(requestedLogLevel)
+		}
+		fmt.Fprintf(os.Stderr, "invalid value %q for %s\n", reqLogLevelStr, EnvTanzuCLILogLevel)
+	}
+	return defaultLogThreshold
+}
+
+// logger defines a logr.Logger
 type logger struct {
 	threshold  *int32
 	level      int32
--- a/log/logger_test.go
+++ b/log/logger_test.go
@@ -1,0 +1,77 @@
+// Copyright 2021 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package log
+
+import (
+	"bytes"
+	"os"
+	"testing"
+
+	"github.com/tj/assert"
+)
+
+func TestLogger(t *testing.T) {
+	assert := assert.New(t)
+
+	tests := []struct {
+		test                  string
+		logLevel              string
+		containStrings        []string
+		doesNotContainStrings []string
+	}{
+		{
+			test:                  "when TANZU_CLI_LOG_LEVEL is not configured",
+			logLevel:              "",
+			containStrings:        []string{"log-default", "log-0", "log-1", "log-3"},
+			doesNotContainStrings: []string{"log-5", "log-8"},
+		},
+		{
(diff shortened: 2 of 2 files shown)


mpanchajanya (approved):
lgtm! Could you add few unit tests for these with env set, unset ?

log/logger.go
```diff
@@ -28,10 +33,19 @@ type logEntry struct {
...
 }
 
-// NewLogger returns a new instance of the clusterctl.
+// NewLogger returns a new instance of the logger.
 func NewLogger() LoggerImpl {
+	logThresold := defaultLogThreshold
+	requestedLogThresoldStr := os.Getenv(EnvTanzuCLILogLevel)
+	if requestedLogThresoldStr != "" {
+		requestedLogThresold, err := strconv.ParseUint(requestedLogThresoldStr, 10, 32)
+		if err == nil {
```
chandrareddyp, line 42: What if user set incorrect value, and we are not capturing the err here, ignoring the error?
IMO, at minimal we need to warn the error message so that user knows that incorrect value has been set.

chandrareddyp (changes requested):
looks good!
Could you please add an example like after setting the env TANZU_CLI_LOG_LEVEL variable, how the UX looks for any command.

log/logger.go
```diff
@@ -28,10 +33,19 @@ type logEntry struct {
 	Values []interface{}
 }
 
-// NewLogger returns a new instance of the clusterctl.
+// NewLogger returns a new instance of the logger.
 func NewLogger() LoggerImpl {
+	logThresold := defaultLogThreshold
```
marckhouzam, line 38: logThresold -> logThreshold

log/logger.go
```diff
@@ -28,10 +33,19 @@ type logEntry struct {
...
+	requestedLogThresoldStr := os.Getenv(EnvTanzuCLILogLevel)
```
marckhouzam, line 39: requestedLogThresoldStr -> requestedLogThresholdStr

log/logger.go
```diff
@@ -28,10 +33,19 @@ type logEntry struct {
...
+	if requestedLogThresoldStr != "" {
+		requestedLogThresold, err := strconv.ParseUint(requestedLogThresoldStr, 10, 32)
```
marckhouzam, line 41: requestedLogThresold -> requestedLogThreshold

log/logger.go
```diff
@@ -8,12 +8,17 @@ import (
...
 	"runtime"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/pkg/errors"
 )
 
+const (
+	EnvTanzuCLILogLevel = "TANZU_CLI_LOG_LEVEL"
```
marckhouzam, line 19: Plugins can use this logging library right?
Should we make the environment variable more generic?

marckhouzam (commented):
Typos and a question

log/logger.go
anujc25: Yes. Considering it is applicable to plugins as well I am using the `TANZU_CLI_LOG_LEVEL` variable. Do you have any other better suggestions?

chandrareddyp (approved):
LGTM!

log/logger.go
marckhouzam: After thinking about a better name, I could not find one. I think what you chose is the best.

marckhouzam (approved):
LGTM
thanks

log/logger_test.go
```diff
@@ -0,0 +1,77 @@
...
+	}{
+		{
+			test:                  "when TANZU_CLI_LOG_LEVEL is not configured",
+			logLevel:              "",
+			containStrings:        []string{"log-default", "log-0", "log-1", "log-3"},
+			doesNotContainStrings: []string{"log-5", "log-8"},
+		},
+		{
+			test:                  "when TANZU_CLI_LOG_LEVEL is set to 3",
+			logLevel:              "",
```
marckhouzam, line 31: This should be a "3" I think

log/logger_test.go
```diff
@@ -0,0 +1,77 @@
...
+			doesNotContainStrings: []string{"log-5", "log-8"},
+		},
+		{
+			test:                  "when TANZU_CLI_LOG_LEVEL is set to 6",
+			logLevel:              "6",
+			containStrings:        []string{"log-default", "log-0", "log-1", "log-3", "log-5"},
+			doesNotContainStrings: []string{"log-8"},
+		},
+		{
+			test:                  "when TANZU_CLI_LOG_LEVEL is set to 9",
```
marckhouzam, line 42: None of the tests verify the actual boundary, meaning setting the level to X and getting logs at X.
I suggest adding such a test or changing the test for level 9 to use level 8.  Or to add a log at level 9.

marckhouzam (commented):
Sorry for having missed this in my previous review.

PR #87, opened 2023-08-02
Allow empty value for config set

chandrareddyp:
This PR is to make sure the `<value>` for the `tanzu config set <key> <value>` is not mandatory, the `<value>` can be empty string.

Fixes #

Unit test cases added.

### Release note

```release-note

```

### DIFF
--- a/config/envs.go
+++ b/config/envs.go
@@ -131,16 +131,10 @@ func SetEnv(key, value string) (err error) {
 	return err
 }
 
-//nolint:dupl
 func setEnv(node *yaml.Node, key, value string) (persist bool, err error) {
 	// check if key is empty
 	if key == "" {
 		return false, errors.New("key cannot be empty")
-	}
-
-	// check if value is empty
-	if value == "" {
-		return false, errors.New("value cannot be empty")
 	}
 
 	// find env node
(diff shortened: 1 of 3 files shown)


config/envs.go
```diff
@@ -42,11 +42,6 @@ func GetEnv(key string) (string, error) {
 }
 
 func getEnv(node *yaml.Node, key string) (string, error) {
-	// check if key is empty
```
mpanchajanya, line 45: @chandrareddyp we need this validation. we dont want key to be empty, only value can be empty

config/envs.go
```diff
@@ -133,11 +128,6 @@ func SetEnv(key, value string) (err error) {
 
 //nolint:dupl
 func setEnv(node *yaml.Node, key, value string) (persist bool, err error) {
-	// check if key is empty
```
mpanchajanya, line 136: @chandrareddyp Same here as well. 

we don't want key to be empty, only value can be empty

mpanchajanya (changes requested):
key cannot be empty. only value can be empty, Try removing validation on value and update the tests as needed

config/envs.go
chandrareddyp: Yes, i raised PR in a rush, i have updated.

mpanchajanya (approved):
lgtm

PR #88, opened 2023-08-15
Update SyncPluginsForTarget API to allow configuring `stdout` and `stderr` externally

anujc25:

The `TanzuPluginSyncForTarget` API is exposed to sync plugins as an experimental API.

* However, when this API is invoked by a plugin, `stdout` and `stderr` are returned as part of the API and are not actively written to the `stdout` or `stderr`. 
* Because of this it makes it feel like the CLI is stuck when installing plugins, especially when there large number of plugin that needs to be installed as part of plugin sync API giving subpar UX to the users.

This change exposes additional `CommandOptions` to configure the `outputWriter` and `errorWriter` using `WithOutputWriter` and `WithErrorWriter` options when invoking the `TanzuPluginSyncForTarget` API, along with writing the output directly to the `os.Stdout` and `os.Stderr` by default.

**Note:** Unlike before, the API will now write the output/error of the plugin sync operation to stderr and stdout by default. Callers who previously manage the output by processing the return values of the API call should consider turning off the default writing to the os.Stdout and os.Stderr by passing WithNoStdout and WithNoStderr options as the additional parameters to the function call in order to avoid duplicate echoing.

Fixes #

* Added unit tests to capture the stdout and stderr along with verifying return types with different combinations

### Release note

```release-note
Fix the `SyncPluginsForTarget` API to allow configuring the `OutputStream` and `ErrorStream` externally. If not configured it defaults the writes to `os.Stdout` and `os.Stderr`
```

### DIFF
--- a/plugin/sync_plugins.go
+++ b/plugin/sync_plugins.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"strings"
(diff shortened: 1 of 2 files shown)


anujc25:
Thanks @marckhouzam  for the review and feedback. I like the `io.MultiWriter` approach.

However, I am not sure we should always write to `os.Stdout` and `os.Stderr` because it is possible that the user does NOT want to print all logs to `os.Stdout` or  `os.Stderr` at all and want to overwrite it with something else.

So I am thinking maybe we can use this `io.MultiWriter` and combine both approaches to give flexibility without changing the API signature.

marckhouzam (commented):
Very nice!  I think this is a very nice API.

We are mixing terminology between `stdout` and any other output stream.
For example `WithStdout` is meant to set a output stream, not necessarily to `stdout`, while `WithNoStdout()` is meant to remove the use of `stdout`.  I recommend a bit of renaming.  I've put suggestions in-line.

plugin/sync_plugins.go
```diff
@@ -43,11 +90,24 @@ func runCommand(commandPath string, args []string) (bytes.Buffer, bytes.Buffer,
...
-func SyncPluginsForTarget(target types.Target) (string, error) {
+// To write the logs to different output and error streams as part of the plugin sync
+// command invocation, configure CommandOptions as part of the parameters.
+// By default this api will write to stdout and stderr.
+//
+// Example:
+//
+//	var outBuf bytes.Buffer
+//	var errBuf bytes.Buffer
+//	SyncPluginsForTarget(types.TargetK8s, WithOutputWriter(outBuf), WithErrorWriter(errBuf))
```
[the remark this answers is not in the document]
marckhouzam: Ok, its just
```suggestion
//	SyncPluginsForTarget(types.TargetK8s, WithOutputWriter(&outBuf), WithErrorWriter(&errBuf))
```

marckhouzam (approved):
One nit about the original example I gave you being wrong but the rest LGTM.
Thanks @anujc25 !

plugin/sync_plugins.go
```diff
@@ -43,11 +90,24 @@ func runCommand(commandPath string, args []string) (bytes.Buffer, bytes.Buffer,
 // implementation are subjected to change/removal if an alternative means to
 // provide equivalent functionality can be introduced.
 //
-// The output of the plugin syncing will be return as a string.
-func SyncPluginsForTarget(target types.Target) (string, error) {
+// To write the logs to different output and error streams as part of the plugin sync
+// command invocation, configure CommandOptions as part of the parameters.
+// By default this api will write to stdout and stderr.
```
vuil, line 95: nit: api -> API. 
The comment block also reads better if this L95 is before L93-94 so the CommandOptions bit leads directly to the example.

vuil (approved):
This is nice enhancement, thanks for the changes. Just some nits on a comment block and the PR description.

On the PR description:

We mention that API outputs to stdout/err by default, but it is not sufficiently clear IMO that this behavior is _new_.

Maybe a tweak to the description would be to augment with something like
Note: Unlike before, the API will now write the output/error of the plugin sync operation to stderr and stdout by default. Callers who previously manage the output by processing the return values of the API call should consider turning off the default writing to the os.Stdout and os.Stderr by passing WithNoStdout and WithNoStderr options as the additional parameters to the function call in order to avoid duplicate echoing.

It wouldn't hurt to make it a tad clearer in the release note as well.  

Finally, given that we mentioned WithNoStdout and WithNoStderr, its seems consistent with describing overriding output and error streams to mention WithOutputWriter and WithErrorWriter as well.

plugin/sync_plugins.go
anujc25: Ack. Updated the comment and PR description.

PR #89, opened 2023-08-18
[release-1.0] Update SyncPluginsForTarget API to allow configuring stdout and stderr externally

anujc25:
Cherry-pick: #88 

This change exposes additional CommandOptions to configure the outputWriter and errorWriter using WithOutputWriter and WithErrorWriter options when invoking the TanzuPluginSyncForTarget API, along with writing the output directly to the os.Stdout and os.Stderr by default.

Note: Unlike before, the API will now write the output/error of the plugin sync operation to stderr and stdout by default. Callers who previously manage the output by processing the return values of the API call should consider turning off the default writing to the os.Stdout and os.Stderr by passing WithNoStdout and WithNoStderr options as the additional parameters to the function call in order to avoid duplicate echoing.

Fixes #

### Release note

```release-note
Fix the `SyncPluginsForTarget` API to allow configuring the `OutputStream` and `ErrorStream` externally. If not configured it defaults the writes to `os.Stdout` and `os.Stderr`
```

### DIFF
--- a/plugin/sync_plugins.go
+++ b/plugin/sync_plugins.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"strings"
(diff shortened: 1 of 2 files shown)


plugin/sync_plugins.go
```diff
@@ -43,11 +90,24 @@ func runCommand(commandPath string, args []string) (bytes.Buffer, bytes.Buffer,
...
-func SyncPluginsForTarget(target types.Target) (string, error) {
+// By default this API will write to os.Stdout and os.Stderr.
+// To write the logs to different output and error streams as part of the plugin sync
+// command invocation, configure CommandOptions as part of the parameters.
+//
+// Example:
+//
+//	var outBuf bytes.Buffer
+//	var errBuf bytes.Buffer
+//	SyncPluginsForTarget(types.TargetK8s, WithOutputWriter(outBuf), WithErrorWriter(errBuf))
```
marckhouzam, line 101: I think the example needs the `&` no?

anujc25: Thanks for catching this. Addressed. I actually created https://github.com/vmware-tanzu/tanzu-plugin-runtime/pull/90 to fix this on the `main` branch as well.

marckhouzam (approved):
LGTM
Thanks!

vuil (approved):
lgtm, thanks

PR #96, opened 2023-09-18
Update CLI command taxonomy word list

vuil:

The global taxonomy wordlist is grossly out-of-date. 
This changes updates it to account of the new terms/flags introduced since the last update.

Note that this change does not immediate affect the lint command output of already-released plugins. 
For a plugin to have more accurately reflection of its lint output, the plugin will have to be rebuilt and linked with the version of the tanzu-plugin-runtime that includes the updated wordlist. 

Also: includes a script `lint-all-plugins.sh` to produce the lint output for all installed plugins.

#### how the word list is updated:

The above-mentioned script was used to produce the terms/flags that are flagged. The output is then audited for any blatant violations, and the rest merged back into the global word list.

Fixes #N/A

`make all`

### Release note

```release-note
Update terms/flags list cli-wordlist.yml to reflect what are in used by plugin commands
```

### DIFF
--- a/plugin/lint/cli-wordlist.yml
+++ b/plugin/lint/cli-wordlist.yml
@@ -1,124 +1,368 @@
 ---
 nouns:
+  - accelerator
   - account
   - admin-kubeconfig
+  - agentartifacts
+  - aks-cluster
   - alpha
   - app
+  - apps
+  - audit
   - available
   - available-upgrades
+  - available-zone
   - aws
+  - backup
+  - backup-location
+  - build-service
+  - builder
+  - buildpack
   - ceip-participation
+  - chart
+  - chartmetadata
+  - claimable
+  - claims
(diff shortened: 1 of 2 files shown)


plugin/lint/cli-wordlist.yml
```diff
@@ -1,147 +1,399 @@
...
+  - audit
   - available
   - available-upgrades
+  - available-zone
   - aws
+  - backup
+  - backup-location
+  - build-service
+  - builder
+  - builder
```
marckhouzam, line 20: Duplicate `builder`

marckhouzam (approved):
LGTM with one nit

plugin/lint/cli-wordlist.yml
vuil: fixed

PR #100, opened 2023-09-25
Add API to get the Kubeconfig for an arbitrary UCP resource

prkalle:
This PR adds API to get the kubeconfig for an arbitrary UCP resource given the UCP context

Summary:
- The new API `ucp.GetKubeconfigForContext()` would construct the kubeconfig for any arbitrary UCP resource of given UCP context. It reads the kubeconfig referred by the UCP context and then updates the kubeconfig's cluster.server url to point to the requested arbitrary resource.
- Client is responsible to provide the valid project/space names. API would not be validating the correctness of project/space names
- API would return error If the context type is not `UCP`.

Fixes #

Unit tests and CI passed

### Release note

```release-note
Add API to get the kubeconfig for an arbitrary UCP resource
```


mpanchajanya:
Also Could you add more details on PR description regarding the changes and new APIs. Thanks

ucp/kubeconfig/kubeconfig.go
```diff
@@ -0,0 +1,81 @@
...
+		Kind:           "Config",
+		APIVersion:     "v1",
+		Clusters:       []*Cluster{{Name: clusterName, Cluster: cluster.Cluster}},
+		AuthInfos:      []*AuthInfo{{Name: userName, AuthInfo: user.AuthInfo}},
+		Contexts:       []*Context{{Name: kubeContextName, Context: context.Context}},
+		CurrentContext: kubeContextName,
+	}, nil
+}
+
+func GetContext(kubeconfig *Config, kubeContextName string) *Context {
```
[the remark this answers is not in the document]
prkalle: > wanted to understand if we are exposing this as a public API to be consumed by cli or plugins? if not lets make it private
Good point. No, these APIs are not meant to be consumed by CLI or plugins. I will move this package to `internal` folder so that methods in `ucp` package can still access these methods.

>naming conflict with existing GetContext API that we have in Configuration APIs ?
I think this shouldn't be an issue as they are part of `kubernetes` package and client would prepend the APIs with package name.

ucp/kubeconfig/kubeconfig.go
```diff
@@ -0,0 +1,81 @@
...
+package kubeconfig
+
+import (
+	"os"
+
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
+)
+
+func ReadKubeConfig(path string) (*Config, error) {
```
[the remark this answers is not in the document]
prkalle: updated. Thanks

ucp/ucp.go
```diff
@@ -0,0 +1,68 @@
...
+
+const (
+	OrgID = "ucpOrgID"
+)
+
+// GetKubeconfigForContext returns the kubeconfig for any arbitrary UCP resource in the UCP object hierarchy
+// referred by the UCP context
+// Pre-reqs: project and space names should be valid
+func GetKubeconfigForContext(contextName, projectName, spaceName string) ([]byte, error) {
+	ctx, err := config.GetContext(contextName)
```
mpanchajanya, line 24: @prkalle Could you few basic null checks for args

prkalle: empty string for `projectName` and `spaceName` are valid inputs and `config.GetContext()` already has check for empty string for `contextName` argument. So we should be good.

mpanchajanya (approved):
lgtm! Thanks for the changes

ucp/ucp.go
anujc25: I saw that we have mentioned that empty strings for `projectName` and `spaceName` are valid inputs. And as per my understanding, it will return the org-level kubeconfig.

However, can we add details about how to get org/project/space level kubeconfig using this API? Providing examples in the `godoc` help above will be really useful for the consumer of the API.

prkalle: Sure. Updated the API `godoc` with examples.

ucp/internal/kubeconfig/types.go
```diff
@@ -0,0 +1,178 @@
...
+		// DisableCompression allows client to opt-out of response compression for all requests to the server. This is useful
+		// to speed up requests (specifically lists) when client-server network bandwidth is ample, by saving time on
+		// compression (server-side) and decompression (client-side): https://github.com/kubernetes/kubernetes/issues/112296.
+		// +optional
+		DisableCompression bool `json:"disable-compression,omitempty" yaml:"disable-compression,omitempty"`
+	} `json:"cluster" yaml:"cluster"`
+}
+
+// AuthInfo contains information that describes identity information.  This is use to tell the kubernetes cluster who you are.
+type AuthInfo struct {
```
vuil, line 70: We can go with this for now, but I would like to explore whether we can process the yaml in a more generic way so we don't have to copy/maintain all this information.

ucp/internal/kubeconfig/kubeconfig.go
```diff
@@ -0,0 +1,86 @@
...
+	err = yaml.Unmarshal(kubeconfig, &config)
+	if err != nil {
+		return nil, err
+	}
+
+	return &config, nil
+}
+
+// MinifyKubeconfig removes all information not used by the provided kubecontext name
+func MinifyKubeconfig(kubeconfig *Config, kubeContextName string) (*Config, error) {
```
vuil, line 30: nit: MinifyKubeConfig to be consistent

vuil (approved):
lgtm, some nits and a general comment about processing the kubeconfig yaml more generically but not blocking.

ucp/internal/kubeconfig/types.go
```diff
@@ -0,0 +1,178 @@
+// Copyright 2023 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package kubeconfig
+
+// Note-IMPORTANT!! The below types are copied from the client-go library and modified a bit(removed fields like extensions not necessary for UCP kubeconfig)
+// to provide kubeconfig access functionality without copying the k8s.io/client-go library which would bring in the other kubernetes dependencies
+// and would impose dependency constraints on the plugin developers using the tanzu-plugin-runtime library
+
+// Config holds the information needed to build connect to remote kubernetes clusters as a given user
```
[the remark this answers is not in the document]
prkalle: This is from the client-go library :) . I fixed it. Thanks

ucp/internal/kubeconfig/kubeconfig.go
prkalle: Updated.

PR #101, opened 2023-09-29
Add context and server sync tests

mpanchajanya:

- Add context and server sync tests
- Tests validating the syncing behaviours between contexts and servers when only context is present and no servers are in config yml and vice versa

Fixes #

make test

### Release note

```release-note
Added context and server sync tests
```


anujc25 (commented):
LGTM. Thanks

anujc25 (approved):

PR #103, opened 2023-09-29
Data as interface in outputwriter

### LINKED ISSUE #102: OutputWriter component does not render container types in YAML/JSON correctly
**Bug description**
The OutputWriter's AddRow() accepts a variatic list of interface{}s, yet immediately converts
each of them to their golang string representation. As a result, if one value is say, a map, it is wrongly rendered for YAML/JSON (in the sense that it is almost certainly not in a format that any requestor of these outputs is expecting to process. )

**Expected behavior**
render these collection types in their respective YAML/JSON implementation

```
  var mapValue = map[string]string{
      "f": "foo",
      "b": "bar",
  }

  var b bytes.Buffer
  tab := NewOutputWriter(&b, string(JSONOutputType), "a", "b", "c")
  tab.AddRow("1", mapValue, 2)
  tab.Render()

  output := b.String()
  fmt.Printf("%v\n", output)
```

expects

```
  [
    {
      "a": "1",
      "b": {
        "b": "bar",
        "f": "foo"
      },
      "c": 2
    }
  ]`
```

**Steps to reproduce the bug / Relevant debug output**
do above, instead of expected output, we get the following today:

```
[
  {
    "a": "1",
    "b": "map[b:bar f:foo]",
    "c": "2"
  }
]
```

similar situations occur with the YAML output

**Plugin Runtime version (also obtainable from `tanzu <pluginName> info`)**
1.0.1

**Environment where the bug was observed (cloud, OS, etc)**

vuil:
```
    Store outputwriter data as 2d array of interfaces

    ... instead of 2d array of strings done prior to this commit.

    AddRow accepts a variatic list of interface{}s, yet immediately converts
    each of them to their golang string representation.

    This change retains these values as interface{}'s instead.
    The idea is to delay any necessary conversion to the point when we are
    actually rendering the values.

    This is particular useful when rendering container values like
    maps/lists in their valid YAML/JSON forms.
```

Fixes #102

make test

### Release note

```release-note
OutputWriter's JSON/YAML outputs non-string tabular fields correctly
```

### DIFF
--- a/component/output.go
+++ b/component/output.go
@@ -44,7 +44,7 @@ const (
 type outputwriter struct {
 	out          io.Writer
 	keys         []string
-	values       [][]string
+	values       [][]interface{}
 	outputFormat OutputType
 }
 
@@ -67,12 +67,8 @@ func (ow *outputwriter) SetKeys(headerKeys ...string) {
 
 // AddRow appends a new row to our table.
 func (ow *outputwriter) AddRow(items ...interface{}) {
-	row := []string{}
-
-	// Make sure all values are ultimately strings
-	for _, item := range items {
-		row = append(row, fmt.Sprintf("%v", item))
-	}
+	row := []interface{}{}
+	row = append(row, items...)
 	ow.values = append(ow.values, row)
 }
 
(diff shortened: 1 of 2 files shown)


prkalle (commented):
lgtm. Thanks.

vuil:
force merge will linter fix is still getting worked on.

PR #113, opened 2023-10-04
Make kubernetes and application-engine type current contexts mutual exclusive

### LINKED ISSUE #114: Make kubernetes and application-engine type current contexts mutual exclusive
The application-engine and kubernetes contexts types are similar in nature because `application-engine` would also be pointing to k8s cluster. Since they both are similar in nature, it is decided to have either of the context types to be active. This would have impact on context based plugins that would be installed or uninstalled based on the current context.

**Additional context**

prkalle:
This PR makes kubernetes and application-engine current contexts mutual exclusive, which means if the k8s context is set as current(active) the TAE context would be removed and vice-versa.

Fixes #114 

Unit tests and CI passed.

### Release note

```release-note
Make kubernetes and application-engine type current contexts mutual exclusive
```

### DIFF
--- a/config/contexts.go
+++ b/config/contexts.go
@@ -394,6 +394,14 @@ func setCurrentContext(node *yaml.Node, ctx *configtypes.Context) (persist bool,
 		currentContextNode.Content = append(currentContextNode.Content, nodeutils.CreateScalarNode(string(ctx.Target), ctx.Name)...)
 		persist = true
 	}
+	// maintain mutual exclusive behavior among all the current context types except TMC
+	// (i.e. there can only be one active current context among all the context types except TMC.
+	//  TMC context type can still be active when other context types are active)
+	if persist {
+		if err := updateMutualExclusiveCurrentContexts(node, ctx); err != nil {
+			return persist, err
+		}
+	}
 	return persist, err
 }
 
@@ -441,5 +449,34 @@ func removeContext(node *yaml.Node, name string) error {
 		contexts = append(contexts, contextNode)
 	}
 	contextsNode.Content = contexts
+	return nil
+}
+
+// updateMutualExclusiveCurrentContexts updates the current contexts to maintain
+// mutual exclusive behavior among the current context types except TMC
+func updateMutualExclusiveCurrentContexts(node *yaml.Node, ctx *configtypes.Context) error {
+	if ctx.Target == configtypes.TargetTMC {
+		return nil
+	}
+
+	cfg, err := convertNodeToClientConfig(node)
+	if err != nil {
+		return err
+	}
+	// deactivate all the other existing current contexts that are not TMC
+	for target, contextName := range cfg.CurrentContext {
+		if target == ctx.Target || target == configtypes.TargetTMC {
+			continue
+		}
+
+		err = removeCurrentContext(node, &configtypes.Context{Target: target})
+		if err != nil {
+			return err
+		}
+		err = removeCurrentServer(node, contextName)
+		if err != nil {
+			return err
+		}
+	}
 	return nil
 }
--- a/config/contexts_test.go
+++ b/config/contexts_test.go
@@ -6,12 +6,20 @@ package config
 import (
 	"testing"
 
+	. "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"gopkg.in/yaml.v3"
 
 	configtypes "github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
 )
+
+func TestCliCmdSuite(t *testing.T) {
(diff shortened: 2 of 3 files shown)


prkalle:
> We need to add Compatibility tests for this change as well.

I will add the Compatibility tests in a follow up PR.

config/contexts_test.go
```diff
@@ -1284,3 +1263,103 @@ func TestSetContextWithEmptyName(t *testing.T) {
...
+
+	Context("tmc context as current context", func() {
+		It("should set tmc context as current context successfully", func() {
+			err = SetCurrentContext("test-tmc")
+			gomega.Expect(err).To(gomega.BeNil())
+			validateCurrentContext(configtypes.TargetTMC, "test-tmc", true, "test-mc-2")
+		})
+	})
+	Context("k8s context as current context", func() {
+		It("should set tmc context as current context successfully", func() {
```
marckhouzam, line 1290: title is wrong

marckhouzam (approved):
Thanks @prkalle !
Two nits, but LGTM

config/contexts.go
```diff
@@ -443,3 +449,33 @@ func removeContext(node *yaml.Node, name string) error {
 	contextsNode.Content = contexts
 	return nil
 }
+
+// updateMutualExclusiveCurrentContexts updates the current contexts to maintain
+// mutual exclusive behavior between K8S and TAE current context types
+func updateMutualExclusiveCurrentContexts(node *yaml.Node, ctx *configtypes.Context) error {
```
vuil, line 455: nit: (I don't feel super strongly about this) I am wondering if we can write this in a more future proof manner by checking for TargetTMC instead.
iow, 
when ctx is not tmc:
   deactivate any ctx that is not tmc

then the code works even when new context types are introduced

config/contexts_test.go
```diff
@@ -6,13 +6,21 @@ package config
 import (
 	"testing"
 
+	. "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
```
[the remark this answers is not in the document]
vuil: +1

config/contexts_test.go
```diff
@@ -1284,3 +1263,103 @@ func TestSetContextWithEmptyName(t *testing.T) {
...
+			validateCurrentContext(configtypes.TargetTMC, "test-tmc", true, "test-mc-2")
+
+			err = SetCurrentContext("test-mc")
+			gomega.Expect(err).To(gomega.BeNil())
+			validateCurrentContext(configtypes.TargetK8s, "test-mc", true, "test-mc")
+
+		})
+	})
+	Context("k8s context as current context after tae context(mutual-exclusion test between k8s and tae) ", func() {
+		It("should have only k8s as current context and tae context should be removed from the curren context", func() {
```
vuil, line 1322: typo curren, same for next It as well.

config/contexts_test.go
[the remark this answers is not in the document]
prkalle: It was not imported into the current package on purpose as the gomega has global variable `Default` which was conflicting with `Default` global variable declared in `config` package.

config/contexts_test.go
prkalle: Thanks, fixed.

config/contexts_test.go
prkalle: fixed.

marckhouzam (commented):
@prkalle There is another `SetCurrentContext()` function in `clientconfig.go`.  It is not clear to me when that one is used, but I get the feeling it can and therefore it may need to be updated as well?

marckhouzam (changes requested):
Blocking the PR until we discuss the last comment https://github.com/vmware-tanzu/tanzu-plugin-runtime/pull/113#pullrequestreview-1664937527

I was just afraid the comment would be missed since I added it after approving

config/contexts.go
prkalle: Sure. Updated the PR to maintain mutual exclusion behavior among all the current context types except TMC. Thanks

prkalle:
> @prkalle There is another `SetCurrentContext()` function in `clientconfig.go`. It is not clear to me when that one is used, but I get the feeling it can and therefore it may need to be updated as well?

`SetCurrentContext()` function in `clientconfig.go` is just to modify the current contexts in the client and user has to call `StoreClientConfig()` method inorder to persist the changes to the config file, which would internally call  `clientConfigSetCurrentContext()` that would endup calling the `setCurrentContext()` in context.go file. So we should be good without updating the `SetCurrentContext()` function in `clientconfig.go`.

cc: @mpanchajanya Thanks for clarification

marckhouzam:
Thanks for the explanation @prkalle (and @mpanchajanya ).  It makes sense.

marckhouzam (approved):
LGTM
Thanks @prkalle !

PR #125, opened 2023-10-16
Remove hardcode of source commit for release notes

vuil:

The hardcode was there during a period when it made sense to include all the release notes for each version released.
Now that 1.0 has been released, release notes should be incremental and relative to the last release.

Fixes #

### Release note

```release-note

```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -59,8 +59,6 @@ jobs:
         fi
         git log -3 --format=oneline
         LAST_TAG="v${major}.${minor}.${revision}"
-        # remove after first release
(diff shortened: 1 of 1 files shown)


marckhouzam (approved):
This change is clearly correct.

However, the algorithm used to compute the `LAST_TAG` does not seem right.
The algorithm subtracts 1 from the last position that is not 0.  So if we want to release `1.1.0`, the `LAST_TAG` will be computed to be `1.0.0`.  However, we have a `1.0.2` release which I believe would be the correct `LAST_TAG`.

I will still approve this PR as a first step.

PR #135, opened 2023-11-07
Compatibility tests to validate k8s and tanzu active context mutual exclusion behavior

prkalle:
This PR addes compatibility tests to validate k8s and tanzu active context mutual exclusion behavior

Fixes #

CI tests passed

### Release note

```release-note

```

### DIFF
--- a/test/compatibility/framework/context/context_options.go
+++ b/test/compatibility/framework/context/context_options.go
@@ -40,6 +40,13 @@ func WithTarget(target types.Target) CfgContextArgsOption {
 func WithContextType(contextType types.ContextType) CfgContextArgsOption {
 	return func(c *CfgContextArgs) {
 		c.ContextType = contextType
+		c.Target = ""
+	}
+}
+
+func WithSetCurrentContext() CfgContextArgsOption {
+	return func(c *CfgContextArgs) {
+		c.SetCurrentContext = true
 	}
 }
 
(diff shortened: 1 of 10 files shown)


test/compatibility/framework/context/context_options.go
```diff
@@ -40,6 +40,13 @@ func WithTarget(target types.Target) CfgContextArgsOption {
 func WithContextType(contextType types.ContextType) CfgContextArgsOption {
 	return func(c *CfgContextArgs) {
 		c.ContextType = contextType
+		c.Target = ""
```
mpanchajanya, line 43: Use WithTarget to set Target

mpanchajanya (changes requested):
looks good. one minor nit

test/compatibility/framework/context/context_options.go
prkalle: I thought of it(using the WithTarget in tests) but I felt like with the latest ContextType, the target would not have relevance(user would leave it to  empty string) in the context tests hence I set Target to empty string whenever user chooses `WithContextType()`. I can change this if you think it would if user explicitly sets them to empty string using `WithTarget`. Let me know your thoughts.

mpanchajanya: makes sense now.  Thanks

mpanchajanya (approved):

PR #138, opened 2023-12-04
Provide a 'LastRefresh' field for OCI discoveries

marckhouzam:

This is meant to be used by the CLI to only refresh the plugin inventory DB at a certain interval, to make the plugin-lifecycle commands more responsive.

I included this field in an OCI discovery so that when we support more than once discovery, we are prepared.

Fixes # N/A

Let CI run.
The rest of the testing is done through the tanzu-cli PR that will use this new field.
I will update this PR with the other PR once it is opened.

### Release note

```release-note
Add `LastRefresh` to an OCI discovery source. 
```

### DIFF
--- a/config/types/clientconfig_types.go
+++ b/config/types/clientconfig_types.go
@@ -234,6 +234,8 @@ type OCIDiscovery struct {
 	// Contains a directory containing YAML files, each of which contains single
 	// CLIPlugin API resource.
 	Image string `json:"image,omitempty" yaml:"image,omitempty"`
+	// LastRefresh allows to store the last time the OCI image was refreshed by the CLI
+	LastRefresh string `json:"lastrefresh,omitempty" yaml:"lastrefresh,omitempty"`
 }
 
 // GenericRESTDiscovery provides a plugin discovery mechanism via any REST API
--- a/config/cli_discovery_sources_it_test.go
+++ b/config/cli_discovery_sources_it_test.go
@@ -79,6 +79,7 @@ currentContext:
     - oci:
         name: default
         image: "/:"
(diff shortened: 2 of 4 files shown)


config/types/clientconfig_types.go
```diff
@@ -234,6 +234,8 @@ type OCIDiscovery struct {
 	// Contains a directory containing YAML files, each of which contains single
 	// CLIPlugin API resource.
 	Image string `json:"image,omitempty" yaml:"image,omitempty"`
+	// LastRefresh allows to store the last time the OCI image was refreshed by the CLI
+	LastRefresh string `json:"lastrefresh,omitempty" yaml:"lastrefresh,omitempty"`
```
vuil, line 238: Wondering if this information 

1. needs to be part of the OCI configuration in the config file.
2. we need a timestamp per OCI discovery source

It feels like something that only the core CLI need to track during it use and the information can be stored somewhere else.

(Setting aside https://github.com/vmware-tanzu/tanzu-cli/pull/605#issuecomment-1847570118 for the moment)
Could storing even a single timestamp in ~/.cache/tanzu/plugin_inventory have done the job as well?

marckhouzam:
After discussing [this comment](https://github.com/vmware-tanzu/tanzu-plugin-runtime/pull/138#discussion_r1420831212) with @vuil, this PR will most probably not needed.

I will mark it as draft until I have updated https://github.com/vmware-tanzu/tanzu-cli/pull/605 and confirmed this can be closed.

marckhouzam:
The [PR that originally needed this change](https://github.com/vmware-tanzu/tanzu-cli/pull/605) was reworked and simplified.
This PR is no longer needed.
Thanks for your good suggestions @vuil

PR #139, opened 2023-12-15
Deprecate IsManagementCluster field

mpanchajanya:

- Deprecate IsManagementCluster field from Context struct

Fixes #

### Release note

```release-note

```

### DIFF
--- a/config/types/clientconfig.go
+++ b/config/types/clientconfig.go
@@ -247,6 +247,7 @@ func (c *ClientConfig) SetActiveContext(contextType ContextType, ctxName string)
 }
 
 // IsManagementCluster tells if the context is for a management cluster.
+// Deprecated: This field is deprecated.  It is currently no used.
 func (c *Context) IsManagementCluster() bool {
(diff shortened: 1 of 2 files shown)


marckhouzam (commented):
@mpanchajanya could explain in the PR description why this can be deprecated?  I still see it being used in tanzu-cli when the user runs `tz context list -o json`

mpanchajanya:
> @mpanchajanya could explain in the PR description why this can be deprecated? I still see it being used in tanzu-cli when the user runs `tz context list -o json`
 
As per @heyjcollins 

```
tanzu context get tap16
name: tap16
target: kubernetes
clusterOpts:
path: /Users/collinsjo/.kube/config
context: tap16
isManagementCluster: true

it is not a Management Cluster.
I assume this boolean is set to true for ALL K8's contexts right now?
We're going to be making all TAP plugins context-scoped in the future.
And will need to be able to represent the TAP cluster contexts accurately.

```

marckhouzam (approved):
LGTM

PR #140, opened 2023-12-15
Update feature flags apis

mpanchajanya:

- Added New Feature APIs
       -  GetAllFeatureFlags
       - ConfigureDefaultFeatureFlags
- Added new ClientConfig methods
       - IsConfigFeatureActivated
       - IsFeatureFlagExists
       - GetAllFeatureFlags 
- Deprecate legacy methods and variables

Fixes #

- Added new tests for the new APIs

### Release note

```release-note

```

### DIFF
--- a/config/features.go
+++ b/config/features.go
@@ -8,10 +8,23 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/config/types"
+
 	"gopkg.in/yaml.v3"
 
 	"github.com/vmware-tanzu/tanzu-plugin-runtime/config/nodeutils"
 )
+
+// GetAllFeatureFlags retrieves all feature flags values from config
+func GetAllFeatureFlags() (map[string]types.FeatureMap, error) {
+	// Retrieve client config node
+	cfg, err := GetClientConfig()
+	if err != nil {
+		return nil, err
+	}
+	return cfg.GetAllFeatureFlags()
+}
 
 // IsFeatureEnabled checks and returns whether specific plugin and key is true
 func IsFeatureEnabled(plugin, key string) (bool, error) {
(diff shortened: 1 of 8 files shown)


config/features.go
```diff
@@ -204,3 +217,21 @@ func IsFeatureActivated(feature string) bool {
 	}
 	return status
 }
+
+// ConfigureDefaultFeatureFlags configures default feature-flags to ClientConfig if missing
+func ConfigureDefaultFeatureFlags(defaultFeatureFlags map[string]bool) error {
```
anujc25, line 222: This function name `ConfigureDefaultFeatureFlags` seems difficult to understand and might confuse the consumer of an API. I think it would be better to have a more generic API here.

What do you think about renaming this to something like `ConfigureFeatureFlags` with an additional option `skipIfExists` (or some better variable name)
```
func ConfigureFeatureFlags(defaultFeatureFlags map[string]bool, skipIfExists bool) error {
```

anujc25 (commented):
LGTM. Just one minor suggestion on API name/signature.

config/features.go
mpanchajanya: Thanks for the suggestion. I agree . Updated the method definition

anujc25 (approved):
LGTM. Thanks

config/legacy_clientconfig_filesystem.go
```diff
@@ -10,19 +10,23 @@ import (
 var (
 	// legacyLocalDirName is the name of the old local directory in which to look for tanzu state. This will be
 	// removed in the future in favor of LocalDirName.
+	// Deprecated: This legacyLocalDirName is deprecated. Use next get config apis
```
marckhouzam, line 13: All the new deprecations in this file are on private functions. Is there a reason we can’t just remove them?

config/types/clientconfig_features.go
```diff
@@ -0,0 +1,46 @@
...
+// GetAllFeatureFlags returns all feature flags from the config file
+func (c *ClientConfig) GetAllFeatureFlags() (map[string]FeatureMap, error) {
+	if c.ClientOptions != nil && c.ClientOptions.Features != nil {
+		return c.ClientOptions.Features, nil
+	}
+	return nil, errors.New("not found")
+}
+
+// IsFeatureFlagExists returns true if the features section in the configuration object contains any value for the plugin.feature combination
+func (c *ClientConfig) IsFeatureFlagExists(plugin, feature string) bool {
```
marckhouzam, line 21: That function name doesn’t sound good in English. How about just FeatureFlagExists? Or IsFeatureFlagSet?

config/legacy_clientconfig_filesystem.go
mpanchajanya: We already have our legacy store marked for deprecation  I have just added these related private methods  as well to be consistent. We would remove these when we remove the legacy store after the deprecation window.

config/types/clientconfig_features.go
mpanchajanya: Updated the method name

marckhouzam (approved):
LGTM
Thanks

PR #142, opened 2023-12-20
Update `tanzu` context APIs to support ClusterGroup

### LINKED ISSUE #143: Update plugin runtime APIs to account for `clustergroup` support

Once clustergroups are transitioned from namespaces to workspaces.

- CLI users need a means to discover and interact with clustergroups they have access to.
- Plugins need API to determine the clustergroup associated with an active tanzu context
- CLI Context-related UX needs account for a context being associated with a clustergroup

**Desired Outcome**
- Today the CLI Context has the ability to associate with any Org/Project/Space accessible to the user. It should be updated to associate with any accessible clustergroup of a project as well.

**Additional context**

anujc25:
- Updates `tanzu` context APIs to support ClusterGroup
- **Updates the `GetKubeconfigForContext` to make it more generalized. Users can pass resource options as `opts` argument.**
```
- func GetKubeconfigForContext(contextName, projectName, spaceName string) ([]byte, error) {
+ func GetKubeconfigForContext(contextName string, opts ...ResourceOptions) ([]byte, error) {
```
Example usage of `GetKubeconfigForContext`:
```
GetKubeconfigForContext("context-name-1", ForProject("test-project"))
GetKubeconfigForContext("context-name-1", ForProject("test-project"), ForClusterGroup("clustergroup1"))
GetKubeconfigForContext("context-name-1", ForProject("test-project"), ForSpace("space1"))
```

- **Updates the `SetTanzuContextActiveResource` to make it more generalized that now takes the `resourceInfo` object.**
```
- func SetTanzuContextActiveResource(contextName, projectName, spaceName string, opts ...CommandOptions) error {
+ func SetTanzuContextActiveResource(contextName string, resourceInfo ResourceInfo, opts ...CommandOptions) error {
```
Example usage of `SetTanzuContextActiveResource`:
```
SetTanzuContextActiveResource("test-context", ResourceInfo{ProjectName: "projectA", SpaceName: "spaceA"})
SetTanzuContextActiveResource("test-context", ResourceInfo{ProjectName: "projectA", ClusterGroupName: "clustergroupB"})
```

Fixes #143

### Release note

```release-note
Add support for `clustergroup` for the `tanzu` context. This updates the API signature for the `GetKubeconfigForContext` and the `SetTanzuContextActiveResource`
```

### DIFF
--- a/config/tanzu_context.go
+++ b/config/tanzu_context.go
@@ -19,9 +19,10 @@ import (
 
 // keys to Context's AdditionalMetadata map
 const (
-	OrgIDKey       = "tanzuOrgID"
-	ProjectNameKey = "tanzuProjectName"
-	SpaceNameKey   = "tanzuSpaceName"
+	OrgIDKey            = "tanzuOrgID"
+	ProjectNameKey      = "tanzuProjectName"
+	SpaceNameKey        = "tanzuSpaceName"
+	ClusterGroupNameKey = "tanzuClusterGroupName"
 )
 
 const (
@@ -39,6 +40,8 @@ type ResourceInfo struct {
 	ProjectName string
 	// SpaceName name of the Space
 	SpaceName string
+	// ClusterGroupName name of the ClusterGroup
+	ClusterGroupName string
 }
 
 // cmdOptions specifies the command options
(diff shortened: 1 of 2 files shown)


config/tanzu_context.go
```diff
@@ -99,30 +102,82 @@ func runCommand(commandPath string, args []string, opts *cmdOptions) (bytes.Buff
...
+// Use Case 4: Get the kubeconfig pointing to Tanzu clustergroup
+// -> projectName        = "PROJECTNAME"
+// -> spaceName          = ""
+// -> clusterGroupName   = "CLUSTERGROUPNAME"
+// ex: kubeconfig's cluster.server URL : https://endpoint/org/orgid/project/<projectName>/clustergroup/<clustergroupName>
 //
-//	ex: kubeconfig's cluster.server URL:  https://endpoint/org/orgid/project/<projectName>/space/<spaceName>
-func GetKubeconfigForContext(contextName, projectName, spaceName string) ([]byte, error) {
+// Note: Specifying `spaceName` and `clusterGroupName` both at the same time is incorrect input.
+func GetKubeconfigForContext(contextName string, opts ...ResourceOptions) ([]byte, error) {
```
anujc25, line 164: This function only supports getting kubeconfig for the `tanzu` context-type and no other context type.

Should we consider renaming this function to be more specific like `GetKubeconfigForTanzuContext`? 
Or
Should we consider supporting the `kubernetes` context-type and make this API more generic? cc @prkalle @vuil @marckhouzam

marckhouzam: Does renaming cause a backwards compatibility problem?

prkalle: Initially these methods were under `tae` package and it was not an issue as user has to call them as `tae.GetKubeconfigForContext`, but later due to renaming the context type to `tanzu` and re-organizing them under `config` package, it is causing confusion. So, i feel it is good to rename to avoid confusion. I believe these changes are breaking ( I see `SetTanzuContextActiveResource()`signature also changed), so we can use this oppurtunity if possible.

prkalle (commented):
The changes LGTM. 
I would wait for others opinions regarding the discussion regarding renaming the method name `GetKubeconfigForContext()`.

config/tanzu_context.go
anujc25: Thanks for the feedback @prkalle and @marckhouzam.

> Does renaming cause a backwards compatibility problem?

I think all the APIs related to the `tanzu` context are experimental APIs and we should be able to make changes to the API at the moment. Also, my existing changes do update the API signature as well.

anujc25:
> I would wait for others opinions regarding the discussion regarding renaming the method name `GetKubeconfigForContext()`.

I am not planning to do any method name change as part of this PR. If we decide, I can create a follow-up PR to rename the method.

config/tanzu_context.go
anujc25: I will keep this thread open to get feedback from @vuil as well. However, I am not planning to do this rename as part of this PR. If we decide, I will create follow-up PR for the rename.

config/tanzu_context_test.go
```diff
@@ -128,10 +128,27 @@ func TestGetKubeconfigForContext(t *testing.T) {
...
+	assert.ErrorContains(t, err, "incorrect resource options provided. Both space and clustergroup are set but only one can be set")
+
 	// Test getting the kubeconfig for an arbitrary Tanzu resource for non Tanzu context
 	nonTanzuCtx, err := GetContext("test-mc")
 	assert.NoError(t, err)
-	_, err = GetKubeconfigForContext(nonTanzuCtx.Name, "project2", "")
+	_, err = GetKubeconfigForContext(nonTanzuCtx.Name, ForProject("project2"))
 	assert.Error(t, err)
 	assert.ErrorContains(t, err, "context must be of type: tanzu")
 }
```
marckhouzam, line 154: It would be nice to add a test in ` TestGetTanzuContextActiveResource` to make sure the new ClusterGroup is properly returned.

marckhouzam (approved):
LGTM
One nit about adding a extra test.

config/tanzu_context_test.go
anujc25: Done. Thanks for catching this gap.

config/tanzu_context.go
vuil: I'd suggest keeping the name look into add suport the kubernetes context-type in a followup, since the usecase for wanting to use this API for tanzu context is not that different from when context type is kubernetes. I also think it should be quite straightfoward to add this support.

PR #146, opened 2024-01-05
UX: Add emojis

chandrareddyp:
This pull request adds a list of emojis and an API to append a given emoji to text. This change helps highlight the UX by prefixing an emoji to UX logs.

Fixes #

Tested in local, here is the output:
image

### Release note

```release-note

```

### DIFF
--- a/component/colorable-tty.go
+++ b/component/colorable-tty.go
@@ -61,3 +61,8 @@ func BeginsWith(s, prefix string) bool {
 func Green(s string) string {
 	return aurora.Green(s).String()
 }
+
+// PrefixEmoji prefixes a string with an emoji.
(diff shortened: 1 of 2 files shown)


prkalle:
> This seems like a good idea to me but I'd like others to weigh in. cc @anujc25 @vuil @mpanchajanya @prkalle

I like the idea too. I think we should also provide an option to disable emojis/colors if users choose to.

mpanchajanya (commented):
looks good.  one thing pls make sure this is tested with no color env variables being set on terminal
TANZU_CLI_NO_COLOR, NO_COLOR

chandrareddyp:
UX design has changed, hence closing this PR.

PR #148, opened 2024-01-11
UX: Updated the output writer with text format

chandrareddyp:
This pull request updates the output writer to include text format support and replaces `fmt.Fprintln` with `fmt.Fprint` in the spinner output. If the API user wants to add a new line at the end of the text or message, it can be appended.

Fixes #

Unit tests are added
Also manual test has done by adding the text format:
```
❯ ts plugin install setting --target tmc
[i] Refreshing plugin inventory cache for "projects.registry.vmware.com/tanzu_cli/plugins/plugin-inventory:latest", this will take a few seconds.
[i] Reading plugin inventory for "projects.registry.vmware.com/tanzu_cli/plugins/plugin-inventory:latest", this will take a few seconds.
\ Installing plugin 'setting:v0.2.13' with target 'mission-control'
```

### Release note

```release-note

```

### DIFF
--- a/component/output.go
+++ b/component/output.go
@@ -23,6 +23,7 @@ const (
 type OutputWriter interface {
 	SetKeys(headerKeys ...string)
 	AddRow(items ...interface{})
+	SetText(text string)
 	Render()
 }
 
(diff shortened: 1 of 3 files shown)


mpanchajanya (approved):
lgtm

prkalle (commented):
lgtm

marckhouzam (commented):
@chandrareddyp Could you explain why we want this change?  What is the use case for it.

I have concerns I would like to discuss.

PR #150, opened 2024-01-17
Add Context sorter to sort by context name then target

chandrareddyp:
This PR adds a sorter for the Context type to sort contexts, first by context name and then by context type.

Fixes #

1) Unit tests are added
2) Manual test output:
Before fix: Context's are not sorted by name:
image

After fix: Context's are sorted by name:
image

### Release note

```release-note

```

### DIFF
--- a/config/types/clientconfig_sorter.go
+++ b/config/types/clientconfig_sorter.go
@@ -1,0 +1,26 @@
+// Copyright 2024 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package types specifies the types for clientconfig
+package types
+
+// ContextSorter is a type that implements the sort interface for Contexts to sort by name and target.
+type ContextSorter []*Context
+
+// Len returns the length of the ContextSorter.
+func (c ContextSorter) Len() int {
+	return len(c)
+}
+
+// Swap swaps the elements at the given indices.
+func (c ContextSorter) Swap(i, j int) {
+	c[i], c[j] = c[j], c[i]
+}
+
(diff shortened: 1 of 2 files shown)


config/types/clientconfig_sorter_test.go
```diff
@@ -0,0 +1,35 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
```
marckhouzam, line 1: 2024

config/types/clientconfig_sorter.go
```diff
@@ -0,0 +1,26 @@
+// Copyright 2022 VMware, Inc. All Rights Reserved.
```
marckhouzam, line 1: 2024

config/types/clientconfig_sorter.go
```diff
@@ -0,0 +1,26 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+// Package types specifies the types for clientconfig
+package types
+
+// ContextSorter is a type that implements the sort interface for Contexts to sort by name and target.
+type ContextSorter []*Context
```
marckhouzam, line 8: Just to be clear that this `ContextSorter` only allows to sort arrays of *pointers* to `Context`, it will not work with `[]Context`.  But I feel this is ok as it is more efficient to handle pointers.

marckhouzam (approved):
LGTM
Just the copyright dates to fix.

config/types/clientconfig_sorter.go
chandrareddyp: Yes thats correct.

PR #156, opened 2024-01-30
Implement sync threshold values set using env variable TANZU_CLI_LOG_LEVEL

mpanchajanya:
- Logger not picking the latest os level env variables
- Since the logger instance is created once any dynamic env variables set for TANZU_CLI_LOG_LEVEL are not picked by the logger causing desired log levels not printing.

Fixes #

 -Manual testing the tanzu cli with logging custom log levels

```

**Case 1 :-**

In Tanzu CLI
	log.SetVerbosity(4)
	log.V(6).Info("testing logger with 6")


# Log is not printed since verbosity is set lower

mpanchajanya@mpanchajanF09MK ~/v/r/tanzu-cli (bug-fix/cert-update)> tz config get
clientOptions:
    features:
        global:
            context-target-v2: "true"
    env:
        any-plugin: baarr
cli:
    ceipOptIn: "true"
    eulaStatus: accepted
    discoverySources:
        - oci:
            name: default
            image: projects.registry.vmware.com/tanzu_cli/plugins/plugin-inventory:latest
    cliId: 736bddac-be88-4084-a00d-137a4972b487
    telemetry:
        source: /Users/mpanchajanya/.config/tanzu-cli-telemetry/cli_metrics.db
mpanchajanya@mpanchajanF09MK ~/v/r/tanzu-cli (bug-fix/cert-update)> tz version
version: v1.2.0-dev
buildDate: 2024-02-01
sha: 4b232f86-dirty
arch: arm64
mpanchajanya@mpanchajanF09MK ~/v/r/tanzu-cli (bug-fix/cert-update)> 


# Log is printed if TANZU_CLI_LOG_LEVEL is set since the env variable takes precedence

mpanchajanya@mpanchajanF09MK ~/v/r/tanzu-cli (bug-fix/cert-update)> tz config set env.TANZU_CLI_LOG_LEVEL 9
mpanchajanya@mpanchajanF09MK ~/v/r/tanzu-cli (bug-fix/cert-update)> tz version
[i] testing logger with 6
version: v1.2.0-dev
buildDate: 2024-02-01
sha: 4b232f86-dirty
arch: arm64
mpanchajanya@mpanchajanF09MK ~/v/r/tanzu-cli (bug-fix/cert-update)> 


**Case 2:-**

# In Tanzu CLI
	log.SetVerbosity(9)
	log.V(6).Info("testing logger with 6")

# Log is printed since verbosity is set higher

mpanchajanya@mpanchajanF09MK ~/v/r/tanzu-cli (bug-fix/cert-update)> tz version
[i] testing logger with 6
version: v1.2.0-dev
buildDate: 2024-02-01
sha: 4b232f86-dirty
arch: arm64
mpanchajanya@mpanchajanF09MK ~/v/r/tanzu-cli (bug-fix/cert-update)> tz config get
[i] testing logger with 6
clientOptions:
    features:
        global:
            context-target-v2: "true"
    env:
        any-plugin: baarr
cli:
    ceipOptIn: "true"
    eulaStatus: accepted
    discoverySources:
        - oci:
            name: default
            image: projects.registry.vmware.com/tanzu_cli/plugins/plugin-inventory:latest
    cliId: 736bddac-be88-4084-a00d-137a4972b487
    telemetry:
        source: /Users/mpanchajanya/.config/tanzu-cli-telemetry/cli_metrics.db
mpanchajanya@mpanchajanF09MK ~/v/r/tanzu-cli (bug-fix/cert-update)> 

# Log is not printed since TANZU_CLI_LOG_LEVEL is set to lower since the env variable takes precedence

mpanchajanya@mpanchajanF09MK ~/v/r/tanzu-cli (bug-fix/cert-update)> tz config set env.TANZU_CLI_LOG_LEVEL 4
[i] testing logger with 6
mpanchajanya@mpanchajanF09MK ~/v/r/tanzu-cli (bug-fix/cert-update)> tz version
version: v1.2.0-dev
buildDate: 2024-02-01
sha: 4b232f86-dirty
arch: arm64
mpanchajanya@mpanchajanF09MK ~/v/r/tanzu-cli (bug-fix/cert-update)> 


```

### Release note

```release-note

```

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -81,87 +81,76 @@ type LoggerImpl interface {
 	WithCallDepth(callDepth int) LoggerImpl
 }
 
+var l = NewLogger()
+
 // Info logs a non-error message with the given key/value pairs as context.
 func Info(msg string, kvs ...interface{}) {
-	l := NewLogger()
 	l.Print(msg, nil, string(LogTypeINFO), kvs...)
 }
 
 // Infof logs a non-error message with the given key/value pairs as context.
 func Infof(format string, args ...interface{}) {
-	l := NewLogger()
 	msg := fmt.Sprintf(format, args...)
 	l.Print(msg, nil, string(LogTypeINFO))
 }
 
 // Error logs an error message with the given key/value pairs as context.
 func Error(err error, msg string, kvs ...interface{}) {
-	l := NewLogger()
 	l.Print(msg, err, string(LogTypeERROR), kvs...)
 }
 
 // Errorf logs a error message with the given key/value pairs as context.
 func Errorf(format string, args ...interface{}) {
-	l := NewLogger()
 	msg := fmt.Sprintf(format, args...)
 	l.Print(msg, nil, string(LogTypeERROR))
 }
 
 // Warning logs a warning messages with the given key/value pairs as context.
 func Warning(msg string, kvs ...interface{}) {
-	l := NewLogger()
 	l.Print(msg, nil, string(LogTypeWARN), kvs...)
 }
 
 // Warningf logs a warning messages with the given message format with format specifier and arguments.
 func Warningf(format string, args ...interface{}) {
-	l := NewLogger()
 	msg := fmt.Sprintf(format, args...)
 	l.Print(msg, nil, string(LogTypeWARN))
 }
 
 // Success logs a success messages with the given key/value pairs as context.
 func Success(msg string, kvs ...interface{}) {
-	l := NewLogger()
 	l.Print(msg, nil, string(LogTypeSUCCESS), kvs...)
 }
 
 // Successf logs a success messages with the given message format with format specifier and arguments.
 func Successf(format string, args ...interface{}) {
-	l := NewLogger()
 	msg := fmt.Sprintf(format, args...)
 	l.Print(msg, nil, string(LogTypeSUCCESS))
 }
 
 // Fatal logs a fatal message with the given key/value pairs as context and returns with os.Exit(1)
 func Fatal(err error, msg string, kvs ...interface{}) {
-	l := NewLogger()
 	l.Print(msg, err, string(LogTypeERROR), kvs...)
 	os.Exit(1)
 }
 
 // Outputf writes a message to stdout
 func Outputf(format string, args ...interface{}) {
-	l := NewLogger()
 	msg := fmt.Sprintf(format, args...)
 	l.Print(msg, nil, string(LogTypeOUTPUT))
 }
 
 // V returns an InfoLogger value for a specific verbosity level.
 func V(level int) LoggerImpl {
-	l := NewLogger()
 	return l.CloneWithLevel(level)
 }
 
 // WithName adds a new element to the logger's name.
 func WithName(name string) LoggerImpl {
-	l := NewLogger()
 	return l.Clone().WithName(name)
 }
 
 // WithValues adds some key-value pairs of context to a logger.
 func WithValues(kvList ...interface{}) LoggerImpl {
-	l := NewLogger()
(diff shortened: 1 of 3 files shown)


log/log.go
```diff
@@ -182,6 +193,7 @@ func ShowTimestamp(show bool) {
 
 // SetVerbosity sets verbosity level and also updates default verbosity level
 func SetVerbosity(verbosity int32) {
+	l := NewLogger()
```
marckhouzam, line 196: I believe all the above changes are good except this one.
The call to `l.SetThreshold(&verbosity)` actually modifies the threshold of the global logger, as is mentioned in the comment above.

This means that if I call `log.SetVerbosity(9)` then all other calls to the logger should respect this new threshold.  You can try it before and after this PR.
If you add the following lines to, say, the `tanzu init` command:
```
log.SetVerbosity(9)
log.V(6).Info("testing logger with 6")
```
and then run `tanzu init` you will see the printout because of the new verbosity (==threshold).
But if you try that same thing with the PR, the printout will no longer show because the verbosity was set on a "temporary" logger.

log/logger.go
```diff
@@ -41,16 +41,39 @@ func NewLogger() LoggerImpl {
...
+
+type LogOptions func(l *LogConfig)
+
+func WithThreshold(threshold int32) LogOptions {
+	return func(l *LogConfig) {
+		l.Threshold = threshold
+	}
+}
+
+func getLogThreshold(opts ...LogOptions) int32 {
```
marckhouzam, line 57: Since the options are not meant to be used externally, can’t we just pass a `defaultThreshold` parameter to this function and avoid the properties pattern?  Or does that pattern provide some thing else?

log/log.go
mpanchajanya: Good catch. Thanks, I have updated the implementation.

log/logger.go
mpanchajanya: makes sense, updated

log/logger.go
```diff
@@ -35,22 +35,24 @@ type logEntry struct {
...
 func NewLogger() LoggerImpl {
-	logThreshold := getLogThreshold()
+	logThreshold := getLogThreshold(defaultLogThreshold)
 	return &logger{
 		threshold: &logThreshold,
 	}
 }
 
-func getLogThreshold() int32 {
+func getLogThreshold(threshold int32) int32 {
```
marckhouzam, line 44: How about naming the parameter `defaultThreshold`?  
I find that a function named `getThreshold()` receiving a parameter named `threshold` is pretty confusing, as that is normally the pattern for `setThreshold()`

log/logger.go
```diff
@@ -65,6 +67,20 @@ type logger struct {
...
+	if l.threshold != nil {
+		logThreshold = getLogThreshold(*l.threshold)
+	} else {
+		logThreshold = getLogThreshold(defaultLogThreshold)
+	}
+	l.threshold = &logThreshold
+}
+
 // SetThreshold implements a New Option that allows to set the threshold level for a logger.
 // The logger will write only log messages with a level/V(x) equal or higher to the threshold.
```
marckhouzam, line 85: Since we're here, isn't this comment "...equal or higher to the threshold." wrong?
It should be "equal or lower" no?

marckhouzam (approved):
Two small things to fix but after that LGTM
Thanks for this fix!

log/logger.go
mpanchajanya: Agree. Updated the args

log/logger.go
mpanchajanya: Yes, Corrected the comment

PR #157, opened 2024-02-02
Refactor Spinner component and Add `SetText` and `StartSpinner` APIs

anujc25:

- Refactors Spinner component to eliminate use of parameters and uses `OutputWriterSpinnerOptions` to create Spinner
- Add `SetText` and `StartSpinner` APIs

Fixes #

- Manual Testing

### Release note

```release-note
`NewOutputWriterSpinnerWithOptions` is being deprecated in favor of the more flexible `NewOutputWriterSpinner` implementation.

`NewOutputWriterSpinner` now supports `SetText` and `StartSpinner` APIs.
```

### DIFF
--- a/component/README.md
+++ b/component/README.md
@@ -177,42 +177,75 @@ This will output:
 {"Age":30,"Name":"John"}
 ```
 
-## OutputWriterSpinner Component
+## Spinner Component
 
-`OutputWriterSpinner` provides an interface to `OutputWriter` augmented with a spinner. It allows for rendering output with a spinner while also providing the ability to stop the spinner and render the final output.
+`Spinner` provides an interface to `OutputWriter` augmented with a spinner. It allows for rendering output of multiple types (Text, Table, Yaml, Json) with a spinner while also providing the ability to stop the spinner and render the final output.
 
 ### Usage
 
-To use `OutputWriterSpinner`, you can import the package in your Go code and create an instance of the `OutputWriterSpinner` interface using the `NewOutputWriterSpinnerWithOptions` function.
+To use `Spinner`, you can import the package in your Go code and create an instance of the `Spinner` interface using the `NewSpinner` function.
 
-Note that NewOutputWriterWithSpinner has been deprecated in favor of NewOutputWriterSpinnerWithOptions
-One difference in the default rendering behavior for YAML/JSON output between
-the writer created with NewOutputWriterWithSpinner vs that created with NewOutputWriterSpinnerWithOptions
-is that row fields no longer gets auto stringified in the latter.
+Note that NewOutputWriterWithSpinner and NewOutputWriterSpinnerWithOptions has been deprecated in favor of NewSpinner
+The main difference is how the parameters are provided when creating a spinner component.
 
-To retain the stringify behavior (unlikely what if needed except for backward compatibility
-reasons), create the output writer by including WithAutoStringify() in the
-options list .
+#### Using Spinner to display Table/JSON/YAML output
 
 ``` go
 import "github.com/vmware-tanzu/tanzu-plugin-runtime/component"
 
 
-// create new OutputWriterSpinner
-outputWriterSpinner, err := component.NewOutputWriterSpinnerWithOptions(os.Stdout, "json", "Loading...", true, []OutputWriterOption{})
+// Create new Spinner component
+spinner, err := component.NewSpinner(component.WithOutputStream(os.Stderr),
+		component.WithOutputFormat(component.TableOutputType), // For JSON use JSONOutputType and for YAML use YAMLOutputType
+		component.WithSpinnerText("Fetching data..."),
+		component.WithSpinnerStarted(),
+        component.WithHeaders("Namespace", "Name", "Ready"))
 if err != nil {
-    fmt.Println("Error creating OutputWriterSpinner:", err)
+    fmt.Println("Error creating spinner:", err)
     return
 }
 
-// Render output with spinner
-outputWriterSpinner.RenderWithSpinner()
+// Do some processing to fetch the data from server and fill rows for table
+spinner.AddRow("default", "pod1", "False")
+spinner.AddRow("default", "pod2", "True")
 
-// Stop spinner and render final output
-outputWriterSpinner.StopSpinner()
+// Render output to stop the spinner and render output in table format
+spinner.Render()
+
 ```
 
-The `NewOutputWriterWithSpinner` function takes in the following parameters:
+#### Using Spinner to display plain text
+
+``` go
+import "github.com/vmware-tanzu/tanzu-plugin-runtime/component"
+
+
+// Create new Spinner component
+spinner, err := component.NewSpinner()
+if err != nil {
+    fmt.Println("Error creating spinner:", err)
+    return
+}
+
+spinner.SetText("Installing plugin 'apps'")
+spinner.StartSpinner()
+
+// Do some processing to fetch and install the required plugin
+err := InstallPlugin("apps")
+
+if err != nil {
+    spinner.SetFinalText("Error while installing the plugin 'apps'", log.LogTypeERROR)
+} else {
+    spinner.SetFinalText("Successfully installed the plugin 'apps'", log.LogTypeSuccess)
+}
+
+// Stop the spinner and render/output FinalText
+spinner.StopSpinner()
+
+```
+
+
+The `NewSpinner` function takes in the following optional options:
 
 - `output io.Writer`: The output writer for the spinner and final output.
 - `outputFormat string`: The output format for the final output. It can be either "json" or "yaml".
--- a/component/README.md
+++ b/component/README.md
@@ -193,8 +193,8 @@ The main difference is how the parameters are provided when creating a `OutputWr
 ``` go
 import "github.com/vmware-tanzu/tanzu-plugin-runtime/component"
 
-// Create new OutputWriterSpinner component
-owSpinner, err := component.NewOutputWriterSpinner(component.WithOutputStream(os.Stderr),
+// Create new OutputWriterSpinner component. If `WithOutputStream` option is not provided, it will use os.Stdout as default output stream
+owSpinner, err := component.NewOutputWriterSpinner(
         component.WithOutputFormat(component.TableOutputType), // For JSON use JSONOutputType and for YAML use YAMLOutputType
         component.WithSpinnerText("Fetching data..."),
         component.WithSpinnerStarted(),
@@ -219,7 +219,7 @@ owSpinner.Render()
 import "github.com/vmware-tanzu/tanzu-plugin-runtime/component"
 
 // Create new OutputWriterSpinner component
-spinner, err := component.NewOutputWriterSpinner()
+spinner, err := component.NewOutputWriterSpinner(component.WithOutputStream(os.Stderr))
 if err != nil {
     fmt.Println("Error creating spinner:", err)
     return
--- a/component/README.md
+++ b/component/README.md
@@ -177,40 +177,39 @@ This will output:
 {"Age":30,"Name":"John"}
 ```
 
-## Spinner Component
+## OutputWriterSpinner Component
 
-`Spinner` provides an interface to `OutputWriter` augmented with a spinner. It allows for rendering output of multiple types (Text, Table, Yaml, Json) with a spinner while also providing the ability to stop the spinner and render the final output.
+`OutputWriterSpinner` provides an interface to `OutputWriter` augmented with a spinner. It allows for rendering output of multiple types (Text, Table, Yaml, Json) with a spinner while also providing the ability to stop the spinner and render the final output.
 
 ### Usage
 
-To use `Spinner`, you can import the package in your Go code and create an instance of the `Spinner` interface using the `NewSpinner` function.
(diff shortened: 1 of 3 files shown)


component/output_spinner.go
```diff
@@ -76,18 +120,16 @@ func NewOutputWriterSpinnerWithOptions(output io.Writer, outputFormat, spinnerTe
 	return setAndInitializeSpinner(ows, spinnerText, startSpinner)
 }
 
-// NewOutputWriterSpinnerWithSpinnerOptions returns implementation of OutputWriterSpinner.
-func NewOutputWriterSpinnerWithSpinnerOptions(output io.Writer, outputFormat OutputType, spinnerText string, startSpinner bool, opts OutputWriterSpinnerOptions, headers ...string) (OutputWriterSpinner, error) {
+// NewSpinner returns implementation of Spinner
+func NewSpinner(opts ...SpinnerOption) (Spinner, error) {
 	ows := &outputwriterspinner{}
-	ows.out = output
```
marckhouzam, line 82: If the caller did not use `WithOutputWriter()`, then `ows.out` will be uninitialized and we will get a panic when we stop the spinner.  Maybe we can set it to `os.Stdout` if it is not set?

Or we could keep the `output io.Writer` as an argument to the constructor

component/output_spinner.go
```diff
@@ -76,18 +120,16 @@ func NewOutputWriterSpinnerWithOptions(output io.Writer, outputFormat, spinnerTe
...
-	ows.outputFormat = outputFormat
```
marckhouzam, line 83: This may also remain uninitialized.  This is not as big a deal as an empty outputFormat will default to Table, but maybe we want to explicitly set it to Table when uninitialized?

marckhouzam (commented):
Very nice improvement.  I think it makes the spinner much more usable.
Thanks a lot for this!

There are a couple of comments that are worth looking at before I approve.

component/output_spinner.go
```diff
@@ -28,25 +35,62 @@ type OutputWriterSpinner interface {
...
+
+// WithOutputFormat sets output format for the spinner component
+func WithOutputFormat(outputFormat string) SpinnerOption {
+	return func(ows *outputwriterspinner) {
+		ows.outputFormat = OutputType(outputFormat)
+	}
+}
+
+// WithOutputWriter sets output writer for the spinner component
+func WithOutputWriter(writer io.Writer) SpinnerOption {
```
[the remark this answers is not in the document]
anujc25: Agreed. I am thinking about renaming this to `WithStdout`.

component/output_spinner.go
anujc25: Good point. I am inclined towards setting it to `os.Stdout` by default. If user wants to update the `out` to something else it can be done with the `WithStdout` option.

component/output_spinner.go
```diff
@@ -76,18 +120,16 @@ func NewOutputWriterSpinnerWithOptions(output io.Writer, outputFormat, spinnerTe
...
-	ows.applyOutputWriterSpinnerOptions(opts.SpinnerOptions)
-	return setAndInitializeSpinner(ows, spinnerText, startSpinner)
+	ows.applySpinnerOptions(opts)
+	return initializeSpinner(ows)
 }
 
 // setAndInitializeSpinner sets the spinner text and initializes the spinner
+//
+// Deprecated: setAndInitializeSpinner because NewOutputWriterSpinnerWithOptions is deprecated
 func setAndInitializeSpinner(ows *outputwriterspinner, spinnerText string, startSpinner bool) (OutputWriterSpinner, error) {
```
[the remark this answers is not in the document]
anujc25: Makes sense. Thanks for the suggestion. I will update it.

component/output_spinner.go
```diff
@@ -28,25 +35,62 @@ type OutputWriterSpinner interface {
...
+// WithHeaders sets key headers
+func WithHeaders(headers ...string) SpinnerOption {
 	return func(ows *outputwriterspinner) {
-		ows.spinnerFinalText = finalText
+		ows.keys = headers
+	}
+}
+
+// WithOutputFormat sets output format for the spinner component
+func WithOutputFormat(outputFormat string) SpinnerOption {
```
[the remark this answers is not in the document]
anujc25: Makes sense. I will update this.

component/output_spinner.go
anujc25: I would prefer not to initialize to anything by default because A user wants to use the spinner for just displaying text without any JSON, YAML or Table outputs. So, that user is never going to use any of the underlying rendering functionality and setting it to table does not feel like the right thing to do (even though if someone trying that it will display the table format because of the default outputFormat).

component/output_spinner.go
```diff
@@ -76,18 +120,16 @@ func NewOutputWriterSpinnerWithOptions(output io.Writer, outputFormat, spinnerTe
 	return setAndInitializeSpinner(ows, spinnerText, startSpinner)
 }
 
-// NewOutputWriterSpinnerWithSpinnerOptions returns implementation of OutputWriterSpinner.
-func NewOutputWriterSpinnerWithSpinnerOptions(output io.Writer, outputFormat OutputType, spinnerText string, startSpinner bool, opts OutputWriterSpinnerOptions, headers ...string) (OutputWriterSpinner, error) {
```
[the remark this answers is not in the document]
anujc25: Thanks for catching this. Updated.

component/output_spinner.go
```diff
@@ -109,16 +151,42 @@ func setAndInitializeSpinner(ows *outputwriterspinner, spinnerText string, start
...
+		attachedToTerminal := isatty.IsTerminal(os.Stdout.Fd()) || isatty.IsCygwinTerminal(os.Stdout.Fd())
+		if ows.startSpinnerOnInit && attachedToTerminal {
+			ows.spinner.Start()
+		}
+	}
+	return ows, nil
+}
+
 // RenderWithSpinner will stop spinner and render the output
 func (ows *outputwriterspinner) RenderWithSpinner() {
```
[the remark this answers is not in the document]
anujc25: Wow. Vui and myself were even discussing this same thing but I thought we might not be able to overwrite the `Render()` functionality and hence never tried this. But the way you mentioned, It seems like it is possible to overwrite the `Rander()` function and I verified after the implementation as well.
Thanks for this suggestion. 👍  Vui and myself did not like the existing naming of `RenderWithSpinner` as well and were thinking about deprecating this functionality.

component/output_spinner.go
```diff
@@ -28,32 +40,69 @@ type OutputWriterSpinner interface {
...
+
+// WithOutputFormat sets output format for the spinner component
+func WithOutputFormat(outputFormat OutputType) SpinnerOption {
+	return func(ows *outputwriterspinner) {
+		ows.outputFormat = outputFormat
+	}
+}
+
+// WithStdout sets stdout for the spinner component
+func WithStdout(writer io.Writer) SpinnerOption {
```
marckhouzam, line 96: We haven’t been good about using the term stdout previously but since this is an API let me bring it up. 

From my understanding “standard out” is the normal place where output goes === `os.Stdout` so it’s confusing to name the function `WithStdout()` since we’re not changing “standard “out but really seeing the output stream for the spinner. How about naming the function `WithOutputStream`? or `WithOutStream`?

component/output_spinner.go
marckhouzam: Good point

component/output_spinner.go
```diff
@@ -65,60 +114,70 @@ func NewOutputWriterWithSpinner(output io.Writer, outputFormat, spinnerText stri
...
 
 // RenderWithSpinner will stop spinner and render the output
+//
+// Deprecated: RenderWithSpinner is being deprecated in favor of Render.
 func (ows *outputwriterspinner) RenderWithSpinner() {
-	if ows.spinner != nil && ows.spinner.Active() {
-		ows.spinner.Stop()
-		fmt.Fprintln(ows.out)
-	}
+	ows.StopSpinner()
```
marckhouzam, line 162: This should be removed since the new render function is calling it already

component/output_spinner.go
marckhouzam: Agreed 👍

marckhouzam (approved):
This is great!  LGTM
2 nits

chandrareddyp (approved):
LGTM.

Can we rename output_spinner.go to spinner.go ?

component/output_spinner.go
anujc25: Ahh. Yes completely agree, I remember you bringing this up earlier as well for a related topic. Thanks for reminding me and suggesting this.

component/output_spinner.go
anujc25: Makes sense. Updated.

component/spinner.go
```diff
@@ -0,0 +1,209 @@
...
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/log"
+)
+
+// OutputWriterSpinner is OutputWriter augmented with a spinner.
+//
+// Deprecated: OutputWriterSpinner is being deprecated in favor of Spinner
+type OutputWriterSpinner Spinner
+
+// Spinner is OutputWriter augmented with a spinner.
+type Spinner interface {
```
vuil, line 24: I think it might still be clearer to use the previous name for the interface because it is not obvious that the Spinner interface encompasses the outputwriter functionality

I would have used a clearer name like OutputWriterWithSpinner, but that a bit too trivial to warrant a deprecation.

anujc25: Ack. Updated the PR to use existing `OutputWriterSpinner` interface along with `NewOutputWriterWithSpinner` as constructor.

component/README.md
```diff
@@ -179,51 +179,88 @@ This will output:
...
+#### Using Spinner to display Table/JSON/YAML output
 
 ``` go
 import "github.com/vmware-tanzu/tanzu-plugin-runtime/component"
 
-
-// create new OutputWriterSpinner
-outputWriterSpinner, err := component.NewOutputWriterSpinnerWithOptions(os.Stdout, "json", "Loading...", true, []OutputWriterOption{})
+// Create new OutputWriterSpinner component
+owSpinner, err := component.NewOutputWriterSpinner(component.WithOutputStream(os.Stderr),
```
marckhouzam, line 197: I realize you want to show the use of `component.WithOutputStream()` but for a section `Using Spinner to display Table/JSON/YAML output`, I think we should use `os.Stdout`.  Maybe you can add a comment that `os.Stdout` is the default, if you want

component/README.md
```diff
@@ -179,51 +179,88 @@ This will output:
...
+
+The created `OutputWriterSpinner` instance provides following methods:
+
+- `SetKeys(headerKeys ...string)`: sets the headers for the OutputWriter component
+- `AddRow(items ...interface{})`: adds rows to the OutputWriter component
+- `SetText(text string)`: sets the spinner text
+- `SetFinalText(finalText string, prefix log.LogType)`: sets the spinner final text and prefix log indicator
+- `StartSpinner()`: start the spinner instance and renders the spinnerText
+- `StopSpinner()`: stops the running spinner instance and renders FinalText if set
+- `Render()`: stops the spinner and render the output
```
marckhouzam, line 263: 👍

component/output_spinner.go
```diff
@@ -65,60 +109,69 @@ func NewOutputWriterWithSpinner(output io.Writer, outputFormat, spinnerText stri
...
+func initializeSpinner(ows *outputwriterspinner) (OutputWriterSpinner, error) {
 	if ows.outputFormat != JSONOutputType && ows.outputFormat != YAMLOutputType {
-		ows.spinnerText = spinnerText
-		ows.spinner = spinner.New(spinner.CharSets[9], 100*time.Millisecond)
+		ows.spinner = spinner.New(spinner.CharSets[9], 100*time.Millisecond, spinner.WithWriter(ows.out))
 		if err := ows.spinner.Color("bgBlack", "bold", "fgWhite"); err != nil {
 			return nil, err
 		}
-		ows.spinner.Suffix = fmt.Sprintf(" %s", spinnerText)
+		ows.spinner.Suffix = fmt.Sprintf(" %s", ows.spinnerText)
```
marckhouzam, line 139: Nit: since you improved things so nicely, why not use `spinner.WithSuffix()` in the constructor?

component/output_spinner.go
```diff
@@ -65,60 +109,69 @@ func NewOutputWriterWithSpinner(output io.Writer, outputFormat, spinnerText stri
...
 		if ows.spinnerFinalText != "" {
 			spinner.WithFinalMSG(ows.spinnerFinalText)(ows.spinner)
```
marckhouzam, line 141: We could also put this one in the constructor.

marckhouzam (commented):
Some final nits

marckhouzam (approved):
LGTM @anujc25 !
The nits are optional.

component/README.md
vuil: Some suggestions on the descriptions for the last 3 items

start the spinner instance and renders the spinnerText ->
start the spinner instance, showing the spinnerText

stops the running spinner instance and renders FinalText if set ->
stops the running spinner instance, displays FinalText if set

stops the spinner and render the output -> stops the running spinner instance, displays FinalText if set, then renders the output

The point is trying to be clear that the display of the spinner text and the render of the accumulated output are two different things (hence only use render for the latter)

anujc25: Updated.

vuil (approved):
Thanks for the updates! This is a nice piece of refactoring that will also make it easier for future enhancements!

PR #162, opened 2024-02-14
Update GetKubeconfigForContext to return kubeconfig for kubernetes co…

mpanchajanya:
…ntext

Fixes #

### Release note

```release-note

```


vuil:
A little more description in the PR description and commit would be useful.
Please update the release notes section as well.

config/contexts_test.go
```diff
@@ -469,6 +469,18 @@ func setupForGetContext() error {
 					Endpoint: "test-endpoint",
 				},
 			},
+			{
+				Name:        "test-k8s",
+				ContextType: configtypes.ContextTypeK8s,
+				GlobalOpts: &configtypes.GlobalServer{
```
vuil, line 475: nit: I don't think contexttype k8s context sets the globalopts, it is harmless to keep this case, but we should definitely have one where globalopts is not set

config/tanzu_context_test.go
```diff
@@ -100,13 +100,25 @@ func TestGetKubeconfigForContext(t *testing.T) {
...
-	c.ClusterOpts.Context = "tanzu-cli-mytanzu"
+	c.ClusterOpts.Context = myTanzu
 	err = SetContext(c, false)
 	assert.NoError(t, err)
 
+	// Setup kubernetes context
+	k8sCtx, err := GetContext("test-k8s")
+	assert.NoError(t, err)
+	k8sCtx.ClusterOpts.Path = kubeconfigFilePath.Name()
+	k8sCtx.ClusterOpts.Context = myTanzu
```
vuil, line 118: nit: this technically not wrong, but in practice it is highly unlikely that a tanzu-type and k8s-type CLI context would share the same kube context. Also, "tanzu-cli-" prefix convention is introduced for tanzu-type context, not what one would typically expect of the k8s type one. Best set up a different kubecontext for the latter.

config/contexts_test.go
mpanchajanya: make sense, updated the test case

config/tanzu_context_test.go
mpanchajanya: Updated the tests with separate k8s kubecontext

anujc25 (commented):
LGTM. 👍

vuil (approved):
lgtm, thanks

config/tanzu_context.go
```diff
@@ -180,10 +180,11 @@ func GetKubeconfigForContext(contextName string, opts ...ResourceOptions) ([]byt
 		opt(rOptions)
 	}
 
-	if ctx.ContextType != configtypes.ContextTypeTanzu {
-		return nil, errors.Errorf("context must be of type: %s", configtypes.ContextTypeTanzu)
+	if ctx.ContextType != configtypes.ContextTypeTanzu && ctx.ContextType != configtypes.ContextTypeK8s {
```
marckhouzam, line 183: Could you update the godoc for this function to match the changes?

marckhouzam (approved):
Besides the godoc comment, LGTM

config/tanzu_context.go
mpanchajanya: Thanks updated

PR #163, opened 2024-02-20
Add new target `operations`/`ops`

anujc25:

This PR adds a new target `operations`/`ops`

Fixes #

### Release note

```release-note

```

### DIFF
--- a/config/types/clientconfig.go
+++ b/config/types/clientconfig.go
@@ -44,6 +44,10 @@ const (
 	// TargetGlobal is used for plugins that are not associated with any target
 	TargetGlobal Target = "global"
 
+	// TargetOperations is used for plugins that does operations for application platforms
+	TargetOperations Target = "operations"
+	targetOperations Target = "ops"
+
 	// TargetUnknown specifies that the target is not currently known
 	TargetUnknown Target = ""
 )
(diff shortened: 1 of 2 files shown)


config/types/clientconfig.go
```diff
@@ -44,6 +44,10 @@ const (
 	// TargetGlobal is used for plugins that are not associated with any target
 	TargetGlobal Target = "global"
 
+	// TargetOperations is used for plugins that does operations for application platforms
```
vuil, line 47: comment needs updating. let me get back to you on how best to characterize these groups of commands

marckhouzam (approved):
Changes look good.
I fell upon this minor one though: I think `hack/scripts/lint-all-plugins.sh` must be updated for the new target

config/types/clientconfig.go
```diff
@@ -44,6 +44,10 @@ const (
...
+	TargetOperations Target = "operations"
+	targetOperations Target = "ops"
```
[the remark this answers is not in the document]
anujc25: > In fact, should we deprecated that variable? It is only used for a deprecated function related to contexts.

Yeah. I also feel that we do not need to update `SupportedTargets` but now thinking we can deprecate `SupportedTargets` as it is used in the deprecated function related to contexts. 

I will update the PR.

config/types/clientconfig.go
anujc25: Updated.

marckhouzam (approved):
LGTM
Thanks!

vuil (commented):
lgtm

PR #164, opened 2024-02-29
Augment PluginDescription with fields used in remapping commands

vuil:

Introduce two additional fields in the PluginDescriptor, InvokedAs and SupportedContexttype, marked experimental are added to support plugin level remapping

Fixes #

### Release note

```release-note
Introduce two experimental fields in the PluginDescriptor, InvokedAs and SupportedContexttype to support plugin-level remapping
```

### DIFF
--- a/plugin/types.go
+++ b/plugin/types.go
@@ -104,4 +104,19 @@ type PluginDescriptor struct {
 
 	// DefaultFeatureFlags is default featureflags to be configured if missing when invoking plugin
 	DefaultFeatureFlags map[string]bool `json:"defaultFeatureFlags,omitempty" yaml:"defaultFeatureFlags,omitempty"`
+
+	// InvokedAs provides a specific mapping to how any command provided by this plugin should be invoked as.
+	// If unset (which is equivalent to setting it to ["<PluginDescriptor.Name>"]), commands will typically be invocable
+	// with the Tanzu CLI using "<PluginDescriptor.Name> <command name> commandargs...."
+	// Can be used to specify additional levels in the command hierarchy via values with space-delimited parts.
+	// e.g. ["operations cluster"] implies plugin's command foo will be invoked by the Tanzu CLI using
+	// 'tanzu operations cluster foo...'
+	// EXPERIMENTAL: subject to change prior to the next official minor release
+	InvokedAs []string `json:"invokedAs,omitempty" yaml:"invokedAs,omitempty"`
+
+	// SupportedContextType specifies one or more ContextType that this plugin will specifically apply to.
+	// When no context of matching type is active, the command tree specified by this plugin should be omitted.
+	// When unset, the plugin does not define any specific opinions on this aspect.
+	// EXPERIMENTAL: subject to change prior to the next official minor release
+	SupportedContextType []types.ContextType `json:"supportedContextType,omitempty" yaml:"supportedContextType,omitempty"`
 }
--- a/plugin/plugin_test.go
+++ b/plugin/plugin_test.go
@@ -62,14 +62,16 @@ func TestAddCommands(t *testing.T) {
 	assert := assert.New(t)
 
 	descriptor := PluginDescriptor{
-		Name:        "Test Plugin",
-		Target:      types.TargetGlobal,
-		Description: "Description of the plugin",
-		Version:     "v1.2.3",
-		BuildSHA:    "cafecafe",
-		Group:       "TestGroup",
-		DocURL:      "https://docs.example.com",
-		Hidden:      false,
+		Name:                 "dummy",
+		Target:               types.TargetK8s,
+		Description:          "Description of the plugin",
+		Version:              "v1.2.3",
+		BuildSHA:             "cafecafe",
+		Group:                "TestGroup",
+		DocURL:               "https://docs.example.com",
+		Hidden:               false,
+		InvokedAs:            []string{"dummy2"},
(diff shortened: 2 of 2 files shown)


anujc25 (approved):
LGTM.

plugin/types.go
```diff
@@ -104,4 +104,19 @@ type PluginDescriptor struct {
 
 	// DefaultFeatureFlags is default featureflags to be configured if missing when invoking plugin
 	DefaultFeatureFlags map[string]bool `json:"defaultFeatureFlags,omitempty" yaml:"defaultFeatureFlags,omitempty"`
+
+	// InvokedAs provides a specific mapping to how any command provided by this plugin should be invoked as.
+	// If unset (which is equivalent to setting it to ["<PluginDescript.Name>"]), commands will typically be invocable
+	// with the Tanzu CLI using "<PluginDescriptor.Name> <command name> commandargs...."
+	// Can be used to specify additional levels in the command hiearchy via values with space-delimited parts.
+	// e.g. ["ops cluster"] implies plugin's command foo will be invoked by the Tanzu CLI using
```
anujc25, line 112: nit: As `ops` is an alias of the `operations` command, it would be better to use `operations` in the example here instead of `ops` just to avoid any confusion.

plugin/types.go
```diff
@@ -104,4 +104,19 @@ type PluginDescriptor struct {
 
 	// DefaultFeatureFlags is default featureflags to be configured if missing when invoking plugin
 	DefaultFeatureFlags map[string]bool `json:"defaultFeatureFlags,omitempty" yaml:"defaultFeatureFlags,omitempty"`
+
+	// InvokedAs provides a specific mapping to how any command provided by this plugin should be invoked as.
+	// If unset (which is equivalent to setting it to ["<PluginDescript.Name>"]), commands will typically be invocable
```
marckhouzam, line 109: `<PluginDescript.Name>` -> `<PluginDescriptor.Name>`

plugin/types.go
```diff
@@ -104,4 +104,19 @@ type PluginDescriptor struct {
...
+	// with the Tanzu CLI using "<PluginDescriptor.Name> <command name> commandargs...."
+	// Can be used to specify additional levels in the command hiearchy via values with space-delimited parts.
```
marckhouzam, line 111: `hierarchy`

marckhouzam, line 111: Are we really asking for “space-delimited”? I see the field is an array of string

plugin/types.go
marckhouzam: This brings up a good point. Do we plan on supporting aliases in the InvokedAs values or must those values be the real name of the command?

Also, what happens if a values in InvokedAs list doesn’t exist?

plugin/types.go
```diff
@@ -104,4 +104,19 @@ type PluginDescriptor struct {
...
+	// e.g. ["ops cluster"] implies plugin's command foo will be invoked by the Tanzu CLI using
+	// 'tanzu ops cluster foo...'
+	// EXPERIMENTAL: subject to change prior to the next official minor release
+	InvokedAs []string `json:"invokedAs,omitempty" yaml:"invokedAs,omitempty"`
+
+	// SupportedContextType specifies one of more ContextType that this plugin will specifically apply to.
```
marckhouzam, line 117: “one of more” -> “one or more”

plugin/types.go
```diff
@@ -104,4 +104,19 @@ type PluginDescriptor struct {
...
+	// When no context of matching type is active, the command tree should be omitted by the CLI if possible.
```
marckhouzam, line 118: How about “the command tree should be” -> “ the command tree specified by this plugin should be”

plugin/types.go
```diff
@@ -104,4 +104,19 @@ type PluginDescriptor struct {
...
+	// When unset, the plugin does not define any specific opinions on this aspect.
+	// EXPERIMENTAL: subject to change prior to the next official minor release
+	SupportedContextType []string `json:"supportedContextType,omitempty" yaml:"supportedContextType,omitempty"`
```
marckhouzam, line 121: Should the type instead be `[]ContextType`?

marckhouzam (commented):
A couple of minor clarifications

plugin/types.go
marckhouzam: Ok, I think I understand better the intention now.  `InvokedAs` is an array to allow the plugin to create more than one command under the CLI.  But to achieve invocation from deeper in the command-tree, that is when we would have a space-separated value.

marckhouzam: To be clear, are we supporting invoking from deeper in the command-tree? 
E.g., `InvokedAs: []string{"config get details"}` which would create a `details` command under `tanzu config get`?

vuil: yes, but entries in this array is to mapping an entire plugin's command tree to a different location, so your example is unlikely. A more realistic use case could be to use this as an alternate means to associate a plugin's commands under a high level grouping like "operations" (or however we plan to evolve the cmdGroup concept.

plugin/types.go
vuil: for simplicity, we only support the real name for now.

> Also, what happens if a values in InvokedAs list doesn’t exist?
This will be explained in the https://github.com/vmware-tanzu/tanzu-cli/pull/687
tldr : that particular remap attempt will fail, but the CLI will likely elect to log and carry on.

marckhouzam (approved):
LGTM
thanks!

PR #169, opened 2024-03-07
Remove the Upper casing of headers for the `listtable` output type  and add padding

anujc25:
- Remove the Upper casing for `listtable` header and add padding

Fixes #

### Release note

```release-note
Remove the Upper casing of headers for the `listtable` output type  and add padding
```

### DIFF
--- a/component/output.go
+++ b/component/output.go
@@ -240,9 +240,9 @@ func renderListTable(ow *outputwriter) {
 			}
 			row = append(row, fmt.Sprintf("%v", data[i]))
 		}
-		headerLabel := strings.ToUpper(header) + ":"
+		headerLabel := header + ":"
 		values := strings.Join(row, ", ")
-		fmt.Fprintf(ow.out, "%-"+strconv.Itoa(headerLength)+"s   %s\n", headerLabel, values)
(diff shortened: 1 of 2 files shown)


marckhouzam (commented):
I totally agree that this change makes the output easier to read.
I'm wondering when `listtable` is used?  Because if it is used by scripts, then this change is not backwards-compatible.

For example, if my script does `tz plugin group search -o listtable | grep GROUP` then my grep would fail after this change because it is case-sensitive.

Is it worth the risk?

anujc25:
> I totally agree that this change makes the output easier to read. I'm wondering when `listtable` is used? Because if it is used by scripts, then this change is not backwards-compatible.
> 
> For example, if my script does `tz plugin group search -o listtable | grep GROUP` then my grep would fail after this change because it is case-sensitive.
> 
> Is it worth the risk?

It is used by some commands for user-readable output. For automation we should be using `yaml` or `json` output so I think we should be fine. (I saw the `package` plugin using that for user-readable output)

marckhouzam (approved):
Ok, if there is no backwards-compatibility concerns, LGTM

PR #170, opened 2024-03-07
Update spinner API to stop all running spinners

chandrareddyp:

This PR updates the spinner API to store all spinners and close any that are still active when the user wants to stop the spinner after the command has completed or been interrupted. This will help ensure all spinners are properly closed.

Use case:

If the user is running `tanzu plugin sync`, it installs all the required plugins for each context. Assume that it has installed a few plugins, and now the user wants to terminate the command with Ctrl+C. In this case, the API should stop the current spinner and also check for any other active spinners and terminate them as well. This new API helps handle these types of use cases.

Fixes #

### Release note

```release-note
The spinner API has been updated to track all active spinners and stop them when the command ends or is terminated by the user.
```

### DIFF
--- a/component/output_spinner.go
+++ b/component/output_spinner.go
@@ -44,6 +44,8 @@ type outputwriterspinner struct {
 // OutputWriterSpinnerOption is an option to configure outputwriterspinner
 type OutputWriterSpinnerOption func(*outputwriterspinner)
 
+var spinners []OutputWriterSpinner
+
 // WithSpinnerFinalText sets the spinner final text and prefix log indicator
 // (log.LogTypeOUTPUT can be used for no prefix)
 func WithSpinnerFinalText(finalText string, prefix log.LogType) OutputWriterSpinnerOption {
@@ -153,7 +155,12 @@ func initializeSpinner(ows *outputwriterspinner) OutputWriterSpinner {
 			ows.spinner.Start()
 		}
 	}
+	storeSpinners(ows)
 	return ows
+}
+
+func storeSpinners(s OutputWriterSpinner) {
+	spinners = append(spinners, s)
 }
 
 // RenderWithSpinner stops the running spinner instance, displays FinalText if set, then renders the output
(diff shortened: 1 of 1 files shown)


component/output_spinner.go
```diff
@@ -171,9 +189,18 @@ func (ows *outputwriterspinner) StartSpinner() {
...
+			if s.GetErrorText() != "" {
+				s.SetFinalText(s.GetErrorText(), log.LogTypeERROR)
+			}
+			if ows.spinner != nil && ows.spinner.Active() {
+				ows.spinner.Stop()
+				if ows.spinnerFinalText != "" {
+					fmt.Fprintln(ows.out)
+				}
+			}
+		}
```
anujc25, line 203: We should not be updating this function. This function is to stop specific spinner and should be kept as it is.

We should implement a separate public API like `StopAllSpinners` (or something similar) which invokes `StopSpinner`.

anujc25:
Can you update the PR title and PR description?

component/output_spinner.go
chandrareddyp: The API consumer may confuse that why should he call `StopAllSpinners` as he started a single spinner! If we give a option `StopSpinner` API then they may use just `StopSpinner` API instead of `StopAllSpinners`.

chandrareddyp:
> Can you update the PR title and PR description?

Updated

component/output_spinner.go
anujc25: @chandrareddyp users will not have access to the `outputwriterspinner` object while handling the interrupt and hence this cannot be even invoked.

TPR does need to provide an API to stop all spinner objects because TPR can keep track of all open spinner and close them when needed. The way it is implemented today, It does not make much sense to close all spinners by invoking some random spinner object's `StopSpinner` API.

component/output_spinner.go
```diff
@@ -30,20 +30,25 @@ type OutputWriterSpinner interface {
 	// SetFinalText sets the spinner final text and prefix
 	// log indicator (log.LogTypeOUTPUT can be used for no prefix)
 	SetFinalText(finalText string, prefix log.LogType)
+	// GetErrorText returns the spinner error text if spinner is terminated or interrupted
+	GetErrorText() string
```
anujc25, line 34: What is the purpose of this new API? I do not see it getting used anywhere in the spinner other than setting the value and getting it back.

chandrareddyp: The intention of the `spinnerErrorText` is to use when spinner is terminated, anyhow `spinnerFinalText` is being used, let me remove it.

component/output_spinner.go
```diff
@@ -173,7 +180,18 @@ func (ows *outputwriterspinner) StartSpinner() {
...
 	if ows.spinner != nil && ows.spinner.Active() {
 		ows.spinner.Stop()
-		fmt.Fprintln(ows.out)
+		if ows.spinnerFinalText != "" {
+			fmt.Fprintln(ows.out)
+		}
+	}
+}
+
+// StopAllSpinners stops all running spinners if any
```
anujc25, line 189: nit: It would be nice if we could improve this comment to specify why and how it should be used when using the spinner.

anujc25 (approved):
LGTM. Thanks.

anujc25:
Please add a release note for this PR as well.

marckhouzam (approved):
The code LGTM but could you put in the PR description what scenario we are trying to address?  I believe it is for interrupts, right?  An example of how this would be used would help.  Currently, it is hard to see the added benefit of this new function.

component/output_spinner.go
chandrareddyp: Updated.

PR #172, opened 2024-03-14
Update SetTanzuContextActiveResource API and GetTanzuContextActiveResource to use Project ID along with Project Name

prkalle:
This PR updates the SetTanzuContextActiveResource() API to accept the Project ID along with Project Name to set a Project as active in the Tanzu context. The GetTanzuContextActiveResource API is also updated to return the Project ID along with Project Name for an active project

Fixes #

Tested with below Unit test and made sure the CLI context is updated with Project ID and Project Name.
```
func TestSetTanzuContextActvResource(t *testing.T) {
 
    os.Setenv("TANZU_BIN", "/Users/pkalle/projects/tanzu-cli/artifacts/darwin/amd64/cli/core/v1.3.0-dev/tanzu-cli-darwin_amd64")
    resInfo := ResourceInfo{
       OrgID: "b1d48027-bb69-4a56-a5b8-e941ef29fa4b",
       ProjectName: "unitel-project-ab",
       ProjectID:   "6d1e2d22-5785-43d7-a6da-1102d7aba83b",
       SpaceName: "store-dev",
    }
    err := SetTanzuContextActiveResource("prem-ucp-intg-lat", resInfo)
    assert.NoError(t, err)
}
```

### Release note

```release-note
Update SetTanzuContextActiveResource API and GetTanzuContextActiveResource to use Project ID along with Project Name
```

### Additional information
This PR depends on  tanzu-cli changes (TODO) to update the `tanzu context update tanzu-active-resource ` command to accept the project-id.

### DIFF
--- a/config/tanzu_context.go
+++ b/config/tanzu_context.go
@@ -22,6 +22,7 @@ import (
 const (
 	OrgIDKey            = "tanzuOrgID"
 	ProjectNameKey      = "tanzuProjectName"
+	ProjectIDKey        = "tanzuProjectID"
 	SpaceNameKey        = "tanzuSpaceName"
 	ClusterGroupNameKey = "tanzuClusterGroupName"
 )
@@ -39,6 +40,8 @@ type ResourceInfo struct {
 	OrgID string
 	// ProjectName name of the Project
 	ProjectName string
+	// ProjectID ID of the Project.
+	ProjectID string
 	// SpaceName name of the Space
 	SpaceName string
 	// ClusterGroupName name of the ClusterGroup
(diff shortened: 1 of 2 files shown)


config/tanzu_context.go
```diff
@@ -268,7 +271,13 @@ func SetTanzuContextActiveResource(contextName string, resourceInfo ResourceInfo
...
 
 	altCommandArgs := []string{customCommandName}
-	args := []string{"context", "update", "tanzu-active-resource", contextName, "--project", resourceInfo.ProjectName}
+	args := []string{"context", "update", "tanzu-active-resource", contextName}
+	if resourceInfo.ProjectName != "" {
+		args = append(args, "--project", resourceInfo.ProjectName)
+	}
+	if resourceInfo.ProjectID != "" {
+		args = append(args, "--project-id", resourceInfo.ProjectID)
+	}
```
anujc25, line 280: Do we expect both to be configured always if one of them is configured? Will it make sense to configure `ProjectName` but not `ProjectID` and vice versa?

anujc25 (approved):
Just one minor question. But otherwise Looks good.

config/tanzu_context.go
prkalle: We expect both to be configured always.

anujc25: I am assuming we will be adding that check in the CLI itself and hence we do not need to add any validations here, is that the correct assumption? Or should we be adding that check here?

prkalle: Yes, you are right, the check would be added in CLI itself. I will be adding TODO in CLI for now(and update it late) as we don't want developers using the main face issues until the plugins using this API are updated.

PR #173, opened 2024-03-21
Update GetKubeconfigForContext API to accept Project ID instead of Project Name

prkalle:

This PR updates `GetKubeconfigForContext` API  logic to use/accept Project ID instead of Project Name

Fixes #

### Release note

```release-note
Update GetKubeconfigForContext API to accept Project ID instead of Project Name
```

### DIFF
--- a/config/tanzu_context.go
+++ b/config/tanzu_context.go
@@ -21,6 +21,7 @@ import (
 // keys to Context's AdditionalMetadata map
 const (
 	OrgIDKey            = "tanzuOrgID"
+	OrgNameKey          = "tanzuOrgName"
 	ProjectNameKey      = "tanzuProjectName"
 	ProjectIDKey        = "tanzuProjectID"
 	SpaceNameKey        = "tanzuSpaceName"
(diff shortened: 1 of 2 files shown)


config/tanzu_context.go
```diff
@@ -21,6 +21,7 @@ import (
 // keys to Context's AdditionalMetadata map
 const (
 	OrgIDKey            = "tanzuOrgID"
+	OrgNameKey          = "tanzuOrgName"
```
marckhouzam, line 24: Why this change?

marckhouzam (commented):
So there is not change in functionality, just a rename of a variable, right?
That means that plugins calling `ForProject()` can use either the project name or project id and it will work, correct?
I'm asking to confirm this change is completely backwards-compatible.

config/tanzu_context.go
prkalle: There is requirement to add Organization name as well in the tanzu context metadata. Since this is minor change I added it as part of this PR.

prkalle:
> So there is not change in functionality, just a rename of a variable, right? That means that plugins calling `ForProject()` can use either the project name or project id and it will work, correct? I'm asking to confirm this change is completely backwards-compatible.

The kubeconfig generated should have projectID as part of the cluster.server URL for UCP beta3. This would work for beta2 if plugins passing `project name` instead of `project id`. However the backward compatibility is not guaranteed as these API changes are meant for beta releases.

marckhouzam (approved):
LGTM
Thanks @prkalle

PR #175, opened 2024-04-12
Experimental: Add support to create the Tanzu Hub client

anujc25:
- Adds support to create authenticated Tanzu Hub client for the specified tanzu context
- The Tanzu Hub client exposes a graphQL client implemented using `github.com/Khan/genqlient/graphql` to query the hub resources with graphQL APIs.
- Also provides a Makefile to help plugin authors initialize and generate the stub for the graphQL queries.

Fixes #

### Release note

```release-note
Experimental: Add support to create the Tanzu Hub client
```


config/tanzu_context.go
```diff
@@ -325,6 +326,12 @@ func GetTanzuContextActiveResource(contextName string) (*ResourceInfo, error) {
 	return activeResourceInfo, nil
 }
 
+// RefreshTanzuContextAccessToken refresh access token associated with the context
+func RefreshTanzuContextAccessToken(contextName string) error {
```
vuil, line 330: It was probably intentionally named RefreshTanzu... instead of GetTanzu.... because we are not retrieving anything, but relying on invoking `get-token` to possibly update the access token, so we are guaranteed to obtain a refreshed one when access the CLI context.

Alternatively, what about the idea of providing GetTanzuContextAccessToken that either processes the get-token result directly or obtains the refreshed token from the CLI context as part of the call?

client/hub/client.go
```diff
@@ -0,0 +1,75 @@
...
+const (
+	EnvTanzuHubGraphQLEndpoint = "TANZU_HUB_GRAPHQL_ENDPOINT"
+)
+
+// HubClient client to talk to tanzu hub through graphql apis
+// It includes authenticated graphql client from github.com/Khan/genqlient/graphql
+// that can be used to do graphql queries
+type HubClient struct {
+	ContextName     string
+	GraphQLClientV1 graphql.Client
```
vuil, line 26: What is the V1 referring to?

client/hub/client.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/config"
+)
+
+const (
+	EnvTanzuHubGraphQLEndpoint = "TANZU_HUB_GRAPHQL_ENDPOINT"
+)
+
+// HubClient client to talk to tanzu hub through graphql apis
```
vuil, line 21: nit: tanzu hub -> Tanzu Hub, graphql -> GraphQL 
and in other places as well. (e.g. L55)

vuil:
Have you thought about where instructions/tooling pertaining to generation of query stubs from Hub schema should reside.
I am not sure if they fit in plugin-tooling.mk, but some convenient make targets might be worth considering

config/tanzu_context.go
anujc25: Updated to expose `GetTanzuContextAccessToken` API

client/hub/client.go
anujc25: Removed `v1`.

client/hub/README.md
```diff
@@ -0,0 +1,35 @@
+# Tanzu Hub Client
+
+This package provides plugin authors to create authenticated Tanzu Hub clients that will allow
```
vuil, line 3: This package enables the creation of authenticated Tanzu Hub clients which in turn enables the interaction with Tanzu Hub endpoint through GraphQL queries.

client/hub/README.md
```diff
@@ -0,0 +1,35 @@
...
+them to implement plugins that want to invoke graphQL queries against the Tanzu Hub endpoint for
+the `tanzu` context type.
+
+## Creating a Tanzu Hub Client
+
+To create a Tanzu Hub client, plugin authors can just use the `CreateHubClient(contextName string)` API
```
vuil, line 9:
```suggestion
To create a Tanzu Hub client, use the `CreateHubClient(contextName string)` API
```

client/hub/README.md
```diff
@@ -0,0 +1,35 @@
...
+by providing the `tanzu` context name. An authenticated Tanzu Hub client for the specified tanzu context will be returned.
+This client includes an authenticated GraphQLClient from the `github.com/Khan/genqlient`
```
vuil, line 11:
```suggestion
This client includes an authenticated GraphQLClient from the `github.com/Khan/genqlient` package
```

client/hub/README.md
```diff
@@ -0,0 +1,35 @@
...
+that can be used to do GraphQL queries. Internally it configures the client with a CSP access token for each request.
```
vuil, line 12: a CSP access token -> an access token

do we have plans to auto-refresh the token? Given the CLI single command invocation use case (assuming we are not dealing with very long-running operations), this might not be too relevant. 
If we are open to the possibility, it might be worth mentioning here and add a TODO in client.go

Also worth mentioning (especially for folks jumping to this doc mostly to learn about hub client integration) that the client if authenticated is assured at least 10-15m access to the graphQL endpoint.

anujc25: Updated.

hack/hub/tanzuhub.mk
```diff
@@ -0,0 +1,23 @@
...
+TANZU_HUB_PKG_DIR=$(TANZU_HUB_PKG_BASE_DIR)/hub
+TANZU_HUB_INIT_FILES := queries.graphql main.go genqlient.yaml
+TANZU_HUB_INIT_PKG_URL ?= https://raw.githubusercontent.com/vmware-tanzu/tanzu-plugin-runtime/main/hack/hub/pkg
+TANZU_HUB_SCHEMA_FILE_URL ?=
+
+.PHONY: tanzu-hub-stub-init
+tanzu-hub-stub-init: # Initialize a Tanzu Hub package with stub, schema.graphql and queries.graphql to generate graphQL client APIs
+	mkdir -p $(TANZU_HUB_PKG_DIR)
+	for filename in $(TANZU_HUB_INIT_FILES); do \
+		[ -f $(TANZU_HUB_PKG_DIR)/$$filename ] && echo "Skipping $(TANZU_HUB_PKG_DIR)/$$filename (already exists)" || wget -O "$(TANZU_HUB_PKG_DIR)/$(filename)" $(TANZU_HUB_INIT_PKG_URL)/$$filename ; \
```
anujc25, line 16:
```suggestion
		[ -f $(TANZU_HUB_PKG_DIR)/$$filename ] && echo "Skipping $(TANZU_HUB_PKG_DIR)/$$filename (already exists)" || wget -O "$(TANZU_HUB_PKG_DIR)/$$filename" $(TANZU_HUB_INIT_PKG_URL)/$$filename ; \
```

hack/hub/tanzuhub.mk
```diff
@@ -0,0 +1,23 @@
...
+	done
+	wget -O "$(TANZU_HUB_PKG_DIR)/schema.graphql" $(TANZU_HUB_SCHEMA_FILE_URL)
```
marckhouzam, line 18: This file is the schema.  I assume it should not be modified by the local project?

1. should it be checked-in to git?  mine is 98k lines
2. can this schema change?
3. should a new version be fetched each time the code is compiled?  Or should the project dev run `make tanzu-hub-stub-init` once in a while to refresh?  or do we not need to re-download at all?

hack/hub/tanzuhub.mk
```diff
@@ -0,0 +1,23 @@
...
+TANZU_HUB_PKG_BASE_DIR ?= $(ROOT_DIR)/pkg
+TANZU_HUB_PKG_DIR=$(TANZU_HUB_PKG_BASE_DIR)/hub
+TANZU_HUB_INIT_FILES := queries.graphql main.go genqlient.yaml
+TANZU_HUB_INIT_PKG_URL ?= https://raw.githubusercontent.com/vmware-tanzu/tanzu-plugin-runtime/main/hack/hub/pkg
+TANZU_HUB_SCHEMA_FILE_URL ?=
+
+.PHONY: tanzu-hub-stub-init
+tanzu-hub-stub-init: # Initialize a Tanzu Hub package with stub, schema.graphql and queries.graphql to generate graphQL client APIs
+	mkdir -p $(TANZU_HUB_PKG_DIR)
+	for filename in $(TANZU_HUB_INIT_FILES); do \
```
marckhouzam, line 15: So these three files are stubs but once created, the user may modify then so we don't want to overwrite them, correct?  Or is it only `queries.graphql` that we need to "protect"?  Could the user want to regenerate the other files to update them to a new TPR release?  In such a case, how about putting the files that can be changed under a `generated/` directory that the user can delete at any time and then run `make tanzu-hub-stub-init` again?

marckhouzam (commented):
This is really cool!  Nice way to help project use the client.
Some questions to help me confirm I understood things properly.

config/tanzu_context.go
```diff
@@ -26,6 +27,9 @@ const (
 	ProjectIDKey        = "tanzuProjectID"
 	SpaceNameKey        = "tanzuSpaceName"
 	ClusterGroupNameKey = "tanzuClusterGroupName"
+
+	TanzuMissionControlEndpointKey = "tanzuMissionControlEndpoint"
```
marckhouzam, line 31: just curious what this one is for as it is not used anywhere

client/hub/README.md
```diff
@@ -0,0 +1,38 @@
...
+There are many golang client libraries for the graphQL, however, Tanzu Plugin Runtime uses `github.com/Khan/genqlient` and
+also returns the corresponding graphQL client as part of the Tanzu Hub client creation.
+
+github.com/Khan/genqlient is a Go library to easily generate type-safe code to query a GraphQL API.
+
+To help plugin authors generate the stub for the Tanzu Hub endpoint, a [tanzuhub.mk](../../hack/hub/tanzuhub.mk) has been provided.
+This makefile provides an easy means for the plugin authors to initialize a `hub` package and also generate the stub from the graphQL queries.
+To use this library plugin authors can follow below steps:
+
+1. Import [tanzuhub.mk](../../hack/hub/tanzuhub.mk) to your project's `Makefile`
```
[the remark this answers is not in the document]
anujc25: Yes. I will update this line to specify exactly how to do this.

config/tanzu_context.go
```diff
@@ -332,6 +336,25 @@ func GetTanzuContextActiveResource(contextName string) (*ResourceInfo, error) {
...
+	if err != nil {
+		return "", err
+	}
+
+	tzCtx, err := GetContext(contextName)
+	if err != nil {
+		return "", err
+	}
+
+	if tzCtx == nil || tzCtx.GlobalOpts == nil {
```
marckhouzam, line 351: Should we also check if `tzCtx.GlobalOpts.Auth == nil`, just to make sure we don't make a plugin panic?

marckhouzam (approved):
Ok, this is great!
I'll let you decide what you want to do about the comments as I just started playing with this.

LGTM whatever you decide.

hack/hub/tanzuhub.mk
anujc25: 1. It does not harm to check-in that file in the plugin repository as plugins can manage how it should be consumed and when it needs to be updated. But as it can change, I don't think we should check it into this repository.
2. Yes it can change.
3. The expectation at the moment is to run the `make tanzu-hub-stub-init` to refresh the schema file.

config/tanzu_context.go
anujc25: It is used by the Tanzu Core CLI and can be used by some `operations` plugins to get the TMC endpoint from the context.

config/tanzu_context.go
anujc25: `tzCtx.GlobalOpts.Auth` is not a pointer variable, so we cannot compare it with `nil` right? Also because it is not a pointer, we can expect `tzCtx.GlobalOpts.Auth` to be available if `tzCtx.GlobalOpts == nil`.

Let me know if I am missing something.

hack/hub/tanzuhub.mk
anujc25: I will need time to think about this a bit more. Considering this is an experimental feature, we should be okay to do this change later as well.

config/tanzu_context.go
marckhouzam: Aha. Sorry. I didn’t check the type and just saw a dot access 😊

PR #177, opened 2024-04-16
Extend plugin mapping to command-level remapping

vuil:

Replaces InvokedAs field in the PluginDescriptor with a list of CommandMapEntry's that can provide additional mapping details. In particular it can specify a destination command path of the Tanzu CLI to map to a specific non root command path in the plugin.

Adds support for capturing the invocation "context" by consuming environment variables that are expected to be set by the Tanzu CLI when invoking a command that is being mapped.

Information in this context allows the plugin to construct proper Help information, and learn if the its command being invoked is done via a command-level mapping or not.

Updates the usage to provides better output when help is invoked on a command-level mapping.

Some additional handling added to propagate aliases and description of a mapped subcommand. 

```
    The aliases and description of a remapped subcommand CommandMapEntry
    should default to the Aliases and Short values of the Command in the
    plugin. Hence, the 'info' command will interrogate the plugin command
    tree for these values unless they are already explicitly specified in
    the CommandMapEntry.
```

Fixes #

- Updated unit tests
- Build sample plugins with command-level map and verified that the command's Aliases and Short description are propagated into the elevate Tanzu CLI commands.

### Release note

```release-note

```

### DIFF
--- a/plugin/info.go
+++ b/plugin/info.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"runtime"
 	"runtime/debug"
+	"strings"
 
 	"github.com/spf13/cobra"
 )
@@ -38,6 +39,7 @@ func newInfoCmd(desc *PluginDescriptor) *cobra.Command {
 		Short:  "Plugin info",
 		Hidden: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			updateCommandLevelMapEntries(cmd, desc)
 			pi := pluginInfo{
 				PluginDescriptor:     *desc,
 				PluginRuntimeVersion: getPluginRuntimeVersion(),
@@ -53,6 +55,42 @@ func newInfoCmd(desc *PluginDescriptor) *cobra.Command {
 	}
 
 	return cmd
+}
+
+func getSourceCommandByPath(rootCmd *cobra.Command, cmdPath string) *cobra.Command {
+	hierarchy := strings.Fields(cmdPath)
+	cmd, _, err := rootCmd.Find(hierarchy)
+	if err == nil {
+		return cmd
+	}
+	return nil
+}
+
+func updateCommandLevelMapEntries(cmd *cobra.Command, desc *PluginDescriptor) {
+	if desc == nil || len(desc.CommandMap) == 0 {
+		return
+	}
+
+	rootCmd := cmd.Parent()
+
+	for i := range desc.CommandMap {
+		mapEntry := &desc.CommandMap[i]
+		// not subcommand mapping
+		if mapEntry.SourceCommandPath == "" {
+			continue
+		}
+
+		srcCmd := getSourceCommandByPath(rootCmd, mapEntry.SourceCommandPath)
+		if srcCmd != nil {
+			if len(mapEntry.Aliases) == 0 {
+				mapEntry.Aliases = srcCmd.Aliases
+			}
+
+			if mapEntry.Description == "" {
+				mapEntry.Description = srcCmd.Short
+			}
+		}
+	}
 }
 
 func getPluginRuntimeVersion() string {
--- a/plugin/invoke_context.go
+++ b/plugin/invoke_context.go
@@ -1,0 +1,52 @@
+// Copyright 2024 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package plugin
+
+import (
+	"os"
+	"strings"
+)
+
+// InvocationContext provides details regarding how a plugin's command is being
+// called by the Tanzu CLI. These details allow the plugin to, among other things,
+// construct proper help information, and learn if the command being invoked
+// is done via a command-level mapping or not.
+type InvocationContext struct {
+	// invokedGroup is a space-delimited portion of the Tanzu CLI command
+	// invocation between the CLI binary and the command name itself.
+	// Empty when invoking a top-level command, e.g. "tanzu apply".
+	invokedGroup string
+
+	// invokedCommand is the name of the command in a Tanzu CLI command invocation
+	invokedCommand string
+
+	// sourceCommandPath is a space-delimited path relative to the plugin's
+	// root command of the command being invoked.
+	// This value is empty when the CLI command invoked is mapped to the
+	// plugin's root command,
+	sourceCommandPath string
+}
+
+func (ic *InvocationContext) String() string {
+	return strings.TrimSpace(ic.invokedGroup + " " + ic.invokedCommand)
+}
+
+// GetInvocationContext returns information about how a Tanzu CLI command is invoked.
+// Note that at the moment a valid InvocationContext is only returned when the
+// invoked plugin command (or its ancestor) has been remapped
+func GetInvocationContext() *InvocationContext {
+	invokedGroup := os.Getenv("TANZU_CLI_INVOKED_GROUP")
+	invokedCommand := os.Getenv("TANZU_CLI_INVOKED_COMMAND")
+	sourceCommandPath := os.Getenv("TANZU_CLI_COMMAND_MAPPED_FROM")
+
+	if invokedGroup != "" || invokedCommand != "" {
+		return &InvocationContext{
+			invokedGroup:      invokedGroup,
+			invokedCommand:    invokedCommand,
+			sourceCommandPath: sourceCommandPath,
+		}
+	}
+
+	return nil
+}
--- a/plugin/plugin_test.go
+++ b/plugin/plugin_test.go
@@ -72,15 +72,15 @@ func TestAddCommands(t *testing.T) {
 		Hidden:               false,
 		SupportedContextType: []types.ContextType{types.ContextTypeTanzu},
 		CommandMap: []CommandMapEntry{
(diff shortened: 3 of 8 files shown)


plugin/types.go
```diff
@@ -56,6 +56,34 @@ const (
...
+type CommandMapEntry struct {
+	// SourceCommandPath is a space-delimited path to the command relative to
+	// the root Command of this plugin, with the root Command's path being ""
+	SourceCommandPath string `json:"srcPath" yaml:"srcPath"`
+	// DestinationCommandPath is a space-delimited path to the command relative
+	// to the root Command of the Tanzu CLI
+	DestinationCommandPath string `json:"dstPath" yaml:"dstPath"`
+	// By default, the command previously situated at the
+	// DestinationCommandPath of the Tanzu CLI, if exist, will be the one
+	// overridden by this entry. If this mapping attempt in intended to
```
anujc25, line 69:
```suggestion
	// overridden by this entry. If this mapping attempt is intended to
```

plugin/types.go
```diff
@@ -56,6 +56,34 @@ const (
...
+	// override another part of the Tanzu CLI command tree, the override path should be used.
+	Overrides string `json:"overrides" yaml:"overrides"`
```
anujc25, line 71: For a first-time developer trying to use this feature, it might be difficult to understand this field. I feel the comments added should be sufficient but still, I was wondering whether we should explain each field with an example or not.

anujc25 (commented):
Looks good 👍 . Just a few really minor comments.

plugin/types.go
anujc25: nvm, I see that these are described in detail as part of https://github.com/vmware-tanzu/tanzu-cli/pull/736/files PR with documentation.

plugin/usage.go
```diff
@@ -64,22 +64,53 @@ const (
...
+}
+
+func useLineEx(cmd *cobra.Command, ic *InvocationContext) string {
+	// by checking sourceCommandPath we limit the use of the InvocationContext
+	// to only command-level (not plugin level) mapping
+	if ic == nil || ic.sourceCommandPath == "" {
+		return cmd.UseLine()
+	}
+
+	// TODO(vuil) look into still incorporating relevant parts of UseLine into output
```
marckhouzam, line 86: note so we don't forget this as a follow-up

plugin/types.go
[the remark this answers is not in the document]
vuil: (this was from the older PR that I forgot to submit here)

plugin/invoke_context.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func (ic *InvocationContext) String() string {
+	return strings.TrimSpace(ic.invokedGroup + " " + ic.invokedCommand)
+}
+
+func GetInvocationContext() *InvocationContext {
+	invokedGroup := os.Getenv("TANZU_CLI_INVOKED_GROUP")
+	invokedCommand := os.Getenv("TANZU_CLI_INVOKED_COMMAND")
+	sourceCommandPath := os.Getenv("TANZU_CLI_COMMAND_MAPPED_FROM")
```
marckhouzam, line 38: I believe that in https://github.com/vmware-tanzu/tanzu-cli/pull/736, these variables are only set when command-remapping is being performed.  However, from the naming here (`InvocationContext`, the variable names) it is not clear that all this only work for remapped commands; it could just as well be used for any plugins all the time but it is not.

Maybe we should add a comment to make this clear.

plugin/usage.go
marckhouzam, line 1: The aliases need some tweaks in the help.  Specifically, if a plugin sub-command is mapped to a different name.  For example, I mapped `tanzu builder cli` to `tanzu mycli` but then the alias is incorrect:
```
$ tz mycli -h
Build CLIs

Usage:
  tanzu mycli
  tanzu kubernetes mycli

  tanzu mycli [command]
  tanzu kubernetes mycli [command]

Aliases:
  cli, c
[...]

$ tz cli -h
[x] : unknown command "cli" for "tanzu"

Did you mean this?
	mycli
```

marckhouzam (commented):
Very nice.  I like the clear way for a plugin to indicate any mapping.  I also like that there can be multiple mappings and from sub-commands.  Very powerful.

We'll need some release notes for this nice change.

This is pretty complex stuff to think about so I'm not quite done but here are minor comments to start.

plugin/types.go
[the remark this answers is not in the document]
marckhouzam: "Overrides" seems fine to me 👍

plugin/usage.go
marckhouzam, line 1: The final footer of the help also needs to be fixed up.
For example, I mapped `tanzu builder cli` to `tanzu mycli` but then the footer does not reflect it:
```
$ tz mycli -h
Build CLIs

Usage:
  tanzu mycli
  tanzu kubernetes mycli

  tanzu mycli [command]
  tanzu kubernetes mycli [command]

Aliases:
  cli, c

Available Commands:
  add-plugin  Add a plugin to a repository

Flags:
  -h, --help   help for cli

Use "tanzu plugin builder cli [command] --help" for more information about a command.
Use "tanzu kubernetes plugin builder cli [command] --help" for more information about a command.
```

plugin/plugin_test.go
```diff
@@ -70,8 +70,22 @@ func TestAddCommands(t *testing.T) {
 		Group:                "TestGroup",
 		DocURL:               "https://docs.example.com",
 		Hidden:               false,
-		InvokedAs:            []string{"dummy2"},
 		SupportedContextType: []types.ContextType{types.ContextTypeTanzu},
+		CommandMap: []CommandMapEntry{
```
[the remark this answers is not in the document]
vuil: The original test was written mostly to verify that an expect base count of commands is generated with each NewPlugin. The CommandMap additions does do too much, but I would say it still minimally demo its use, and also that its presence did not cause regression to the original test.

plugin/types.go
```diff
@@ -105,18 +133,15 @@ type PluginDescriptor struct {
...
-	// EXPERIMENTAL: subject to change prior to the next official minor release
-	InvokedAs []string `json:"invokedAs,omitempty" yaml:"invokedAs,omitempty"`
-
 	// SupportedContextType specifies one or more ContextType that this plugin will specifically apply to.
 	// When no context of matching type is active, the command tree specified by this plugin should be omitted.
 	// When unset, the plugin does not define any specific opinions on this aspect.
 	// EXPERIMENTAL: subject to change prior to the next official minor release
 	SupportedContextType []types.ContextType `json:"supportedContextType,omitempty" yaml:"supportedContextType,omitempty"`
+
+	// CommandMap specifies one or more CommandMapEntry's and describes how one
```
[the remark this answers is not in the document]
marckhouzam: Makes sense.  I hadn't realized it was on purpose.  You way is better 👍

plugin/usage.go
vuil: Thanks, the issue is with still assuming the name of the plugin is always a valid alias entry (the first item int the alias list), which is no longer true for a mapped command. I will address it in a followup.

plugin/usage.go
```diff
@@ -111,22 +142,20 @@ func formatHelpFooter(cmd *cobra.Command, target types.Target) string {
 
 	footer.WriteString("\n")
 
+	ic := GetInvocationContext()
+	base := "Use"
```
marckhouzam, line 146: This should be `Use "`.  Right now the quote is missing:
```
$ tz mycli -h
Build CLIs

Usage:
  tanzu mycli [command]
[...]

Use tanzu mycli [command] --help" for more information about a command.
```

plugin/usage.go
```diff
@@ -111,22 +142,20 @@ func formatHelpFooter(cmd *cobra.Command, target types.Target) string {
...
+	if !strings.HasPrefix(cmd.CommandPath(), "tanzu ") {
+		base = "Use tanzu"
```
marckhouzam, line 148: Here to there should be a `"` before `tanzu`

plugin/info.go
```diff
@@ -55,6 +57,42 @@ func newInfoCmd(desc *PluginDescriptor) *cobra.Command {
...
+
+		srcCmd := getSourceCommandByPath(rootCmd, mapEntry.SourceCommandPath)
+		if srcCmd != nil {
+			if len(mapEntry.Aliases) == 0 {
+				mapEntry.Aliases = srcCmd.Aliases
+			}
+
+			if mapEntry.Description == "" {
+				mapEntry.Description = srcCmd.Short
+			}
```
marckhouzam, line 91: Nicely done!

marckhouzam (approved):
Once the missing quote x2 is fixed (see comments) and that CI passes, this LGTM.

Nice job @vuil, this is some powerful stuff!

plugin/usage.go
vuil: nice catch! Updated.

plugin/usage.go
marckhouzam: Fixed in #180

plugin/usage.go
marckhouzam: Fixed in #180

PR #179, opened 2024-04-19
Update Spinner API with UnsetFinalText

chandrareddyp:
This PR adds a new Spinner API to unset the spinner's final text if it has already been set.

This new API helps to handle use cases where the final text needs to be unset. For example, if an error message has been set as the final text to display in case the command fails or is terminated, and the command ends up being successful, you may want to unset the error text as the final text and not display any text at the end.

Fixes #

### Release note

```release-note

```


anujc25:
Can you explain why this will be needed for anyone in the PR description?

chandrareddyp:
> Can you explain why this will be needed for anyone in the PR description?

Updated.

PR #180, opened 2024-04-25
Fix usage of commands involved in command-level mapping

vuil:

Among the changes
 - Fix usage output of a descendant command of the command-level mapped command.
 - Leverage the invoked command's .Use as long as it is prefixed by the premap command name
 - Update aliases list to included the mapped command name instead of the premapped one

In addition, provide a set of functions to obtain useful information about the InvocationContext

```
    CLIInvocationString : return base CLI invocation string up to the mapped command
    CLIInvocationStringForCommand : return full CLI invocation string that would have led to the invocation of the provided Command
    MappedSourceCommandPath : returns the mapped command path used as the source of the command mapping
    InvokedGroupPath : returns the command path CLI uses prior to the mapped command
    InvokedCommandName : returns named of the command CLI uses to invoke the mapped command
```

Fixes #

See unit tests updates.

manual tests:
(dpr is mapped command name of command `deeper`)
```
> tz dpr dshout --help
dshout something

Usage:
  tanzu dpr dshout MESSAGE

Aliases:
  dshout, sh, sht

Flags:
  -h, --help           help for dshout
  -v, --value string   value to add to shout
```

```
> tz dpr --help
deeper commands

Usage:
  tanzu dpr (with no additional arguments)

  tanzu dpr [command]

Aliases:
  dpr, deep

Available Commands:
  dshout      dshout something

Flags:
  -h, --help   help for deeper

Use "tanzu dpr [command] --help" for more information about a command.
```

(dshout prints invocation context and invocation string for the dshout command)
```
tz dpr dshout foo
 foo!!
Invocation Context:
&plugin.InvocationContext{invokedGroup:"", invokedCommand:"dpr", sourceCommandPath:"deeper"}
Invocation String:
dpr dshout
```

### Release note

```release-note
Fix usage string and aliases of help for commands involved in command-level mapping
```

### DIFF
--- a/plugin/invoke_context.go
+++ b/plugin/invoke_context.go
@@ -31,7 +31,7 @@ type InvocationContext struct {
 }
 
 func (ic *InvocationContext) CLIInvocationString() string {
-	return strings.TrimSpace(ic.invokedGroup + " " + ic.invokedCommand)
+	return buildInvocationString(ic.invokedGroup, ic.invokedCommand)
 }
 
 func (ic *InvocationContext) MappedSourceCommandPath() string {
@@ -48,14 +48,14 @@ func (ic *InvocationContext) InvokedGroupPath() string {
 
 // CLIInvocationStringForCommand returns the CLI invocation string when cmd was
 // invoked under this invocation context.
-// Invoking "tanzu + (the string returned) as a CLI command would equivalent to
-// the running of cmd with no arguments
+// Invoking "tanzu + (the string returned) as a CLI command would be equivalent to
+// executing cmd with no arguments
 func (ic *InvocationContext) CLIInvocationStringForCommand(cmd *cobra.Command) string {
-	hierarchy, err := hierarchyFromMappedCommand(cmd, ic.sourceCommandPath)
-	if err != nil {
-		return strings.TrimSpace(ic.invokedGroup + " " + ic.invokedCommand)
+	hierarchy, err := hierarchyFromMappedCommandPath(ic.sourceCommandPath, cmd)
+	if err != nil || len(hierarchy) == 0 {
+		return buildInvocationString(ic.invokedGroup, ic.invokedCommand)
 	}
-	return strings.TrimSpace(ic.invokedGroup+" "+ic.invokedCommand) + " " + strings.Join(hierarchy, " ")
+	return buildInvocationString(ic.invokedGroup, ic.invokedCommand, strings.Join(hierarchy, " "))
 }
 
 // GetInvocationContext returns information about how a Tanzu CLI command is
--- a/plugin/invoke_context.go
+++ b/plugin/invoke_context.go
@@ -6,6 +6,8 @@ package plugin
 import (
 	"os"
 	"strings"
+
+	"github.com/spf13/cobra"
 )
 
 // InvocationContext provides details regarding how a plugin's command is being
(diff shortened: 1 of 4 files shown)


plugin/usage.go
```diff
@@ -76,22 +77,100 @@ func buildInvocationString(parts ...string) string {
...
+
+	// we are executing the actual mapped command
+	if len(hierarchy) == 0 {
+		tail, found := strings.CutPrefix(cmd.Use, cmd.Name())
+		// in this case, it is acceptable to leverage command's .Use string if
+		// it is prefixed by the command name as long as the prefix is stripped
+		if found {
+			return ic.CLIInvocationString() + tail
+		}
+		return ic.CLIInvocationString()
```
marckhouzam, line 153: Not sure when this line would hit, but it does not hurt

plugin/usage.go
```diff
@@ -76,22 +77,100 @@ func buildInvocationString(parts ...string) string {
...
+	}
+
+	// if command is deeper than the mapped command, its .Use string is
+	// considered safe to use in replacement of the command name as long as the
+	// former is prefixed by the latter
+	lastCommandName := hierarchy[len(hierarchy)-1]
+	hierarchy[len(hierarchy)-1] = cmd.Use
```
marckhouzam, line 160: I think the above line should be removed as it is in the `if` below

plugin/usage.go
```diff
@@ -76,22 +77,100 @@ func buildInvocationString(parts ...string) string {
...
-	// TODO(vuil) look into still incorporating relevant parts of UseLine into output
-	return ic.String()
+	hierarchy, err := hierarchyFromMappedCommandPath(ic.sourceCommandPath, cmd)
+	if err != nil {
+		return ic.CLIInvocationString()
+	}
+
+	// we are executing the actual mapped command
+	if len(hierarchy) == 0 {
+		tail, found := strings.CutPrefix(cmd.Use, cmd.Name())
```
marckhouzam, line 147: `CutPrefix()` was added in go 1.20 but our runtime only requires go 1.18 (to avoid forcing plugin devs to upgrade go).  This is why CI is failing.

marckhouzam (approved):
Some minor fixes but after that LGTM

plugin/usage.go
vuil: thanks for catching, updated.

plugin/invoke_context.go
```diff
@@ -28,11 +30,36 @@ type InvocationContext struct {
 	sourceCommandPath string
 }
 
-func (ic *InvocationContext) String() string {
+func (ic *InvocationContext) CLIInvocationString() string {
 	return strings.TrimSpace(ic.invokedGroup + " " + ic.invokedCommand)
```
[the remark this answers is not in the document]
vuil: yep, missed that one. thanks.

plugin/usage.go
vuil: when in the very unlikly event there is no prefix match in the .Use, we elect not to use it.

plugin/usage.go
```diff
@@ -163,9 +243,20 @@ func formatHelpFooter(cmd *cobra.Command, target types.Target) string {
 
 func aliasesWithMappedName(cmd *cobra.Command) string {
 	cmdName := cmd.Name()
+	// if root cmd
 	if v, ok := cmd.Annotations[cobra.CommandDisplayNameAnnotation]; ok {
 		cmdName = v
+	} else {
+		ic := GetInvocationContext()
```
marckhouzam, line 250: `ic` could be nil here.
If I add an alias to a sub-command of a plugin, this new path get triggered and panics

vuil: nice catch, updated.

plugin/usage.go
vuil: updated to remove its use

plugin/usage.go
```diff
@@ -76,22 +77,101 @@ func buildInvocationString(parts ...string) string {
...
+	}
+
+	// we are executing the actual mapped command
+	if len(hierarchy) == 0 {
+		found := strings.HasPrefix(cmd.Use, cmd.Name())
+
+		// in this case, it is acceptable to leverage command's .Use string if
+		// it is prefixed by the command name as long as the prefix is stripped
+		if found {
+			tail := strings.TrimLeft(cmd.Use, cmd.Name())
```
marckhouzam, line 152: We may need `TrimPrefix()` here instead? The godoc for `TrimLeft` says: 
```
func strings.TrimLeft(s string, cutset string) string

TrimLeft returns a slice of the string s with all leading Unicode code points contained in cutset removed.

To remove a prefix, use TrimPrefix instead.
```

vuil: nice catch. Tripped me up on more than one occassion already. TrimLeft is the wrong function to use, so I will update, but funny as it turned out, the nature of the inputs and the preceding line meant that it will always end up with the same result as TrimPrefix.

marckhouzam (approved):
Thanks!

PR #182, opened 2024-04-26
(outdated) Update generate-docs to account for command mapping

vuil:

```   
    The approach taken is to recreate the tanzu CLI command tree with only
    the plugin's commands, after accounting for the mapping directives,
    and hand said command tree to cobra's doc generation APIs.
```

Fixes #

ran plugin-binary generate-docs --docs-dir xxx
for sample plugin binaries with and without CommandMap entries

use Tanzu CLI (need to include PR...)'s generate-all-docs command
tanzu generate-all-docs --docs-dir xxx
with the above mentioned sample plugin binaries installed

and verified that a consistent set of md docs are generated

### Release note

```release-note

```

### DIFF
--- a/plugin/generate_docs.go
+++ b/plugin/generate_docs.go
@@ -7,6 +7,7 @@ package plugin
 import (
 	"fmt"
 	"os"
+	"sort"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


vuil:
replaced by #183

PR #183, opened 2024-04-29
Update generate-docs to accounts for command mapping 2

vuil:

```   
    The approach taken is to recreate the tanzu CLI command tree with only
    the plugin's commands, after accounting for the mapping directives,
    and hand said command tree to cobra's doc generation APIs.
```

Note: this superceds #182 (retained temporarily for reference)

* Added new unit tests to generate-docs using a Plugin with various CommandMap configuration
and verifies generated file set and expected content of the files.

* Run `plugin-binary generate-docs --docs-dir xxx` for sample plugin binaries with and without CommandMap entries and verified that a consistent set of md docs are generated

* Use Tanzu CLI (need to include https://github.com/vmware-tanzu/tanzu-cli/pull/752)'s generate-all-docs command
`tanzu generate-all-docs --docs-dir xxx` with the above mentioned sample plugin binaries installed and verified that a consistent set of md docs are generated

### Release note

```release-note
Fix generate-docs for plugins with command mapping directives
```

### DIFF
--- a/plugin/usage.go
+++ b/plugin/usage.go
@@ -82,6 +82,11 @@ func matchOnCommandName(cmd *cobra.Command, value string) bool {
 }
 
 func findSubCommandByHierarchy(cmd *cobra.Command, hierarchy []string, matcher func(*cobra.Command, string) bool) (*cobra.Command, *cobra.Command) {
+	if len(hierarchy) == 0 {
+		parent := cmd.Parent()
+		return cmd, parent
+	}
+
 	childCmds := cmd.Commands()
 	for i := range childCmds {
 		if len(hierarchy) == 1 {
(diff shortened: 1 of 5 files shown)


plugin/usage.go
```diff
@@ -106,10 +111,7 @@ func hierarchyFromMappedCommandPath(mappedCommandPath string, cmd *cobra.Command
 	var fromCmd *cobra.Command
 	var additionalCmdNames []string
 
-	rootCmd := cmd
-	for rootCmd.HasParent() {
-		rootCmd = rootCmd.Parent()
-	}
+	rootCmd := cmd.Root()
```
marckhouzam, line 114: 👍

marckhouzam (approved):
Nice.  One comment about removing the footer, then we're good.

PR #193, opened 2024-06-13
Add per-mapping required context type

vuil:

This provides explicit control at a command map entry level on whether the mapping should take effect conditionally on the type of context that is active.

Fixes #

### Release note

```release-note
CommandMapEntry supports a RequiredContextType list, which controls whether the mapping should take effect based on the type of the active context.  
```


marckhouzam (approved):
LGTM

anujc25 (approved):
LGTM.

PR #198, opened 2024-07-03
Fix OutputWriter Table to show rows when no headers are specified

anujc25:

- Fix OutputWriter Table to show rows when no headers are specified
- This will make sure that we support Tables without headers as well.

Fixes #

- Added unit tests

### Release note

```release-note
Fix OutputWriter Table to show rows when no headers are specified
```


vuil (approved):
lgtm. thanks.

PR #199, opened 2024-07-09
Fix plugin command usage when the `mission-control`/`operations` target plugin remaps the commands 

anujc25:

- This PR fixes the help text generation issue when `non-global` target plugin remaps the commands.

Before this change:
- Target information was shown even if the command gets remapped to the root level

After this change:
- Target information will not be used to show command help text if the command is remapped

Fixes #

- Manual Testing
- Unit Test

### Release note

```release-note
Fix plugin command usage when the `mission-control`/`operations` target plugin remaps the commands 
```


marckhouzam (approved):
LGTM
Thanks for the quick fix!

PR #200, opened 2024-07-17
Update the file locking module dependency

prkalle:
This PR updates the file locking go module dependency

Fixes #

- CI and Unit tests passed.
- Used project plugin and updated the go.mod to point to the local tanzu-plugin-runtime and compiled the project plugin. Later ran the `tanzu project use` command which would update the CLI configuration file (after acquiring the locks using the new library) and also ran the `tanzu context list` command (which is using the old plugin-runtime version without the new locklibrary change) which access the CLI configuration file(after acquiring the locks) in parallel so as to verify the new library locking don't have compatibility issues with the CLI/plugins using the old locking library. Below are the snapshots for the tests and they ran successfully.

Screenshot 2024-07-17 at 11 37 51 AM

Screenshot 2024-07-17 at 11 38 05 AM

### Release note

```release-note

```

### DIFF
--- a/config/clientconfignextgen_lock.go
+++ b/config/clientconfignextgen_lock.go
@@ -10,7 +10,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/juju/fslock"
+	"github.com/alexflint/go-filemutex"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


anujc25 (approved):
LGTM. Thanks

PR #202, opened 2024-08-06
Improvements to Hub client unit testing framework

anujc25:

- Support throwing errors for Query and Mutations
- Make `OperationName` optional when sending requests
- Make mock response more flexible to return by implementing the `Responder` function which includes the mock variables

Fixes #

### Release note

```release-note
Make OperationName optional for Hub Client requests. Improvements to the Hub client unit test framework
```

### DIFF
--- a/client/hub/testing/operations.go
+++ b/client/hub/testing/operations.go
@@ -18,7 +18,11 @@
 
 package testing
 
-import "context"
+import (
+	"context"
+
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/client/hub"
+)
 
 // Operation Type Constants
 const (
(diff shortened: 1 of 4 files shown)


client/hub/testing/operations.go
```diff
@@ -56,13 +60,26 @@ type Operation struct {
...
+	// defining both.
+	Response hub.Response
 
-	// EventGenerator should generate mock events for Subscription operation type
+	// Responder implements the function that based on some parameters should respond
+	// differently. This is mainly for advanced users who want more flexibility in terms
+	// of what the response should look like depending on the request variable provided
+	// a match on Operation.opType, Operation.Name, and Operation.Variables.
+	// Responder is to be used for Query and Mutation operations only.
+	// Note: User can define either `Response` or implement `Responder` function but should be be
```
vuil, line 73: be be -> not be

client/hub/testing/operations.go
```diff
@@ -56,13 +60,26 @@ type Operation struct {
...
-	Response interface{}
+	// Response is to be used for Query and Mutation operations only.
+	// Note: User can define either `Response` or implement `Responder` function but should be be
+	// defining both.
+	Response hub.Response
 
-	// EventGenerator should generate mock events for Subscription operation type
+	// Responder implements the function that based on some parameters should respond
+	// differently. This is mainly for advanced users who want more flexibility in terms
+	// of what the response should look like depending on the request variable provided
```
vuil, line 70: the part after "in terms of" doesn't read quite clearly.
I believe responder is useful if the response is dependent on the Operation processed?

suggestion: Maybe it is sufficient to say something to that effect.
e.g 
"Tests that do not need flexibility in returning different responses based on the Operation should just configure the Response field instead."

client/hub/testing/operations.go
```diff
@@ -95,8 +112,3 @@ type OperationError struct {
 	// as part of the api error whenever the server makes a match on OperationError.Extensions
 	Extensions interface{}
 }
-
-// EventGenerator should implement a eventData generator for testing and
-// send mock event response to the `eventData` channel. To suggest end of
-// the event responses from server side, you can close the eventData channel
-type EventGenerator func(ctx context.Context, eventData chan<- Response)
```
vuil, line 102: This typedef seems useful, any reason to remove it?

vuil (approved):
A very nice extension of the framework capability. Just a new nits.

client/hub/testing/operations.go
anujc25: Updated.

client/hub/testing/operations.go
anujc25: I thought it would not be used by the developers directly but after some thought I feel it still makes sense to keep this separately defined so I have reverted this and also did a similar for the `Responder` type.

PR #203, opened 2024-08-06
Update to graphql testing doc

marckhouzam:

This PR updates the `README` for the graphql testing framework as well as some comments here and there.

Also it fixes two error situations where the mock server would send more than one response because of the use of `break` instead of `return`.

Fixes # N/A

CI

### Release note

```release-note
Update graphql testing README.
```


anujc25 (approved):
LGTM. 🎉

PR #204, opened 2024-08-07
Fix file lock issue on windows

anujc25:

**Background:**
- As part of https://github.com/vmware-tanzu/tanzu-plugin-runtime/pull/200 we switched to use `github.com/alexflint/go-filemutex` package instead of `github.com/juju/fslock`.
- The `Unlock` function between the two libraries behaves differently. The main difference is the `go-filemutex.Unlock` does not close the file whereas the `fslock.Unlock` does close the file, automatically releasing any acquired lock.
- Because of the above difference we noticed that when the following sequence of operations are performed the acquiring the lock fails because the file was kept open by the first lock call.
   - `go-filemutex.Lock`
   - `go-filemutex.Unlock` 
   - `fslock.Lock`
- The above behavior can be noticed when the Tanzu CLI uses `go-filemutex` to lock the file using newer version of tanzu-plugin-runtime and plugins use `fslock` to lock the file using an older version of tanzu-plugin-runtime

**Fix:**
- Instead of using `go-filemutex.Unlock` use `go-filemutex.Close` to unlock the file and close the file descriptor. Doing this will make the newer version of library using `go-filemutex` compatible with older version of library which uses `fslock`

Fixes #

- Manual tests using a custom code that uses both libraries together.
```
// Copyright 2022 VMware, Inc. All Rights Reserved.
// SPDX-License-Identifier: Apache-2.0

package main

import (
	"os"

	"github.com/anujc/test-locking/filemutex"
	"github.com/anujc/test-locking/juju"
)

func main() {
	if len(os.Args) < 2 {
		panic("argument is required. Values: '1' or '2'")
	}

	arg := os.Args[1]
	if arg == "1" {
		filemutex.AcquireLock()
		filemutex.ReleaseLock()

		juju.AcquireLock()
		juju.ReleaseLock()

	} else if arg == "2" {
		juju.AcquireLock()
		juju.ReleaseLock()

		filemutex.AcquireLock()
		filemutex.ReleaseLock()
	} else {
		panic("incorrect argument")
	}
}
```

- Without the fix the above test program results are as follows:
   - Mac & Linux: 
      - `1` -> Success
      - `2` -> Success
   - Windows:
      - `1` -> Failed
      - `2` -> Success

- With the fix the above test program results are as follows:
   - Mac & Linux: 
      - `1` -> Success
      - `2` -> Success
   - Windows:
      - `1` -> Success
      - `2` -> Success

### Release note

```release-note
Fix file lock issue on windows
```

### DIFF
--- a/config/clientconfignextgen_lock.go
+++ b/config/clientconfignextgen_lock.go
@@ -57,7 +57,7 @@ func ReleaseTanzuConfigNextGenLock() {
 	if cfgNextGenLock == nil {
 		return
 	}
-	if errUnlock := cfgNextGenLock.Unlock(); errUnlock != nil {
+	if errUnlock := cfgNextGenLock.Close(); errUnlock != nil {
 		panic(fmt.Sprintf("cannot release lock for tanzu config file, reason: %v", errUnlock))
 	}
 
--- a/config/lock.go
+++ b/config/lock.go
@@ -69,7 +69,7 @@ func ReleaseTanzuConfigLock() {
 	if tanzuConfigLock == nil {
 		return
 	}
-	if errUnlock := tanzuConfigLock.Unlock(); errUnlock != nil {
+	if errUnlock := tanzuConfigLock.Close(); errUnlock != nil {
 		panic(fmt.Sprintf("cannot release lock for tanzu config file, reason: %v", errUnlock))
 	}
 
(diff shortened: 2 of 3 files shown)


marckhouzam (approved):
Awesome! Thank you for this!

vuil (approved):
Thanks for the fix!

PR #206, opened 2024-08-13
The Hub UT framework should send what it received

marckhouzam:

Before this PR the UT framework would send the registered operation.  This was not useful for unit tests as it is those same tests that registered the operation, so they are aware of what it is.  What the tests need is for the framework to send the actual received GraphQL request so that it can be verified. 

This PR does this.
NOTE: This change is not backwards-compatible.  However, I feel it is early enough in the use of the UT framework and that the only ones using it are probably the CLI team, that we can easily adapt.

The unit tests of the UT framework are updated to check that the sent value is the actual query received by the mock server.

Follow-up to #202 

I have an internal set of unit tests that now detects more failures by comparing the received graphql request with what the tests expect.

### Release note

```release-note
Hub unit test framework now sends back the GraphQL request it received for more thorough testing.  This change breaks backwards-compatibility of the API used to write unit test for GraphQL: the `Responder` and `EventGenerator` types have a slightly different signature.
```

### DIFF
--- a/client/hub/client_test.go
+++ b/client/hub/client_test.go
@@ -31,12 +31,21 @@ import (
 //     }
 //   }
 
+// A set of variables that are sent with the query.
+// This allows us to test that the mock server returns these
+// variables in the response.
+var defaultVariables = map[string]interface{}{
+	"name":  "john",
+	"lines": 100,
+}
+
 // getProjects is a wrapper of an `QueryAllProjects“ API call to fetch project names
 // Note: This is only for the testing the MockServer with HubClient
 func getProjects(hc hub.Client) ([]string, error) {
 	req := &hub.Request{
-		OpName: "QueryAllProjects",
-		Query:  QueryAllProjects_Operation,
+		OpName:    "QueryAllProjects",
+		Query:     QueryAllProjects_Operation,
+		Variables: defaultVariables,
 	}
 	var responseData QueryAllProjectsResponse
 	err := hc.Request(context.Background(), req, &responseData)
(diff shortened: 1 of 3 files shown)


client/hub/client_test.go
```diff
@@ -135,21 +198,80 @@ func TestQueryWithTanzuHubClient(t *testing.T) {
...
+					// Note that the variable keys must match the ones in the query, but
+					// the values can be different.
+					Variables: map[string]interface{}{
+						"name":  "notjohn",
+						"lines": 0,
+					},
+					Responder: func(_ context.Context, receivedReq hubtesting.Request) hub.Response {
 						return hub.Response{
-							Errors: gqlerror.List{{Message: fmt.Sprintf("operation %s failed with error %s", op.Identifier, "fake-error-message")}},
+							Errors: gqlerror.List{{Message: fmt.Sprintf("operation failed with error and received %v", receivedReq)}},
```
anujc25, line 212: nit(minor): I am not sure if updating the message as below will help to understand things a bit more or not. Ignore if you feel it does not make sense.

```suggestion
							Errors: gqlerror.List{{Message: fmt.Sprintf("operation failed with error when received request as: %v", receivedReq)}},
```

anujc25 (approved):
LGTM. Thanks for the improvement. This is going to be really useful for writing dynamic unit tests based on request data.

anujc25:
@marckhouzam Can you mention about the breaking change in this unit test framework as part of the ReleaseNotes as well?

PR #210, opened 2024-08-29
Support Self-Signed CA Cert and skip-cert-verify with Hub Client

anujc25:

- The Hub Client will use the `GetCert` API to get the certificate data for the specified hub endpoint.
- If the certificate data is not found, it will skip the TLS configuration and let the client use default configuration.
- GetCert API now accepts URI along with hostname

Fixes #

### Release note

```release-note
Support Self-Signed CA Cert and skip-cert-verify with Hub Client
GetCert API now accepts URI along with hostname
```

### DIFF
--- a/config/certs.go
+++ b/config/certs.go
@@ -26,13 +26,16 @@ func GetCerts() ([]*configtypes.Cert, error) {
 	return getCerts(node)
 }
 
-// GetCert retrieves the cert configuration by host
+// GetCert retrieves the cert configuration by host or URI
 func GetCert(host string) (*configtypes.Cert, error) {
 	if host == "" {
 		return nil, errors.New("host is empty")
 	}
 	u, err := url.Parse(host)
-	if err == nil && u.Hostname() != "" {
+	if err != nil {
+		return nil, err
+	}
+	if u.Hostname() != "" {
 		host = u.Hostname()
 	}
 
@@ -96,7 +99,7 @@ func DeleteCert(host string) error {
 	return persistConfig(node)
 }
 
-// CertExists checks if cert config by host already exists
+// CertExists checks if cert config by host or URI already exists
 func CertExists(host string) (bool, error) {
 	if host == "" {
 		return false, errors.New("host is empty")
--- a/config/certs.go
+++ b/config/certs.go
@@ -5,6 +5,7 @@ package config
 
 import (
 	"fmt"
+	"net/url"
 
 	"github.com/pkg/errors"
 	"gopkg.in/yaml.v3"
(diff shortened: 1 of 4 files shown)


config/certs_test.go
```diff
@@ -43,6 +43,14 @@ func TestSetGetDeleteCerts(t *testing.T) {
 	assert.Nil(t, err)
 	assert.Equal(t, cert1, ctx)
 
+	ctx, err = GetCert("https://test1")
```
prkalle, line 46: nit: This was already existing, but can we change the `ctx` getCert` something so that it is more opt?

prkalle (approved):
LGTM. I have a minor nit comment. Thanks

config/certs_test.go
vuil: +1 
ctx->cert throughout would have made more sense.

config/certs.go
```diff
@@ -30,6 +31,11 @@ func GetCert(host string) (*configtypes.Cert, error) {
 	if host == "" {
```
vuil, line 31: suggestion: update comment to say "by host or URI"

config/certs_test.go
```diff
@@ -43,6 +43,14 @@ func TestSetGetDeleteCerts(t *testing.T) {
...
+	assert.Nil(t, err)
+	assert.Equal(t, cert1, ctx)
+
+	ctx, err = GetCert("https://test1/fake")
```
vuil, line 50: suggestion: add 3 more cases for 
ws://test1/fake and 
wss://test1/fake to show this works for websocket URIs as well.

and 
https://test1:12345/fake

config/certs_test.go
```diff
@@ -43,6 +43,14 @@ func TestSetGetDeleteCerts(t *testing.T) {
...
+	assert.Nil(t, err)
+	assert.Equal(t, cert1, ctx)
+
```
vuil, line 53: nit: suggest add another case of invalid uri 
ctx, err = GetCert("https://tes  t1/fak e") or something, should we choose to return a better error as suggested above.

config/certs.go
```diff
@@ -30,6 +31,11 @@ func GetCert(host string) (*configtypes.Cert, error) {
...
 		return nil, errors.New("host is empty")
 	}
+	u, err := url.Parse(host)
```
vuil, line 34: an unparsable URI ends up getting the same error as a missing entry. 
how about just return here,
with errors.New("invalid uri") or wrap the err?

vuil (commented):
very nice!
Some nits and suggestion for additional test cases.

vuil:
another nit:
Maybe GetCert deserves a mention in docs/config.md?

config/certs_test.go
anujc25: Added this case but I would prefer not to validate the URI/HostName as part of this API and just use the comparison with the existing cert store configuration and return the result.

vuil (approved):
lgtm, thanks

vuil (approved):

PR #212, opened 2024-08-30
Fix incorrect base64 encoded caCert usage with Hub Client

anujc25:
- Fix incorrect base64 encoded caCert usage with Hub Client

Fixes #

### Release note

```release-note
None
```

### DIFF
--- a/client/hub/client.go
+++ b/client/hub/client.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/base64"
 	"net/http"
 	"os"
 	"strconv"
@@ -15,6 +16,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/vmware-tanzu/tanzu-plugin-runtime/config"
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/log"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


client/hub/client.go
```diff
@@ -162,12 +164,22 @@ func (c *hubClient) getTLSConfig() *tls.Config {
 	// If CACertData is present use it
 	if certData.CACertData != "" {
 		var pool *x509.CertPool
-		var err error
+
+		decodedCACertData, err := base64.StdEncoding.DecodeString(certData.CACertData)
+		if err != nil {
+			log.Infof("unable to use custom cert when taking to '%s' endpoint. Error: %s", c.tanzuHubEndpoint, err.Error())
```
vuil, line 170: typo: taking
suggestion maybe just shorten from 'when talking to' -> 'for'

prkalle, line 170: nit: `taking` -> 'talking` or `interacting`(may be the righ one)

client/hub/client.go
```diff
@@ -162,12 +164,22 @@ func (c *hubClient) getTLSConfig() *tls.Config {
...
+		}
+
 		pool, err = x509.SystemCertPool()
 		if err != nil || pool == nil {
 			pool = x509.NewCertPool()
 		}
-		pool.AppendCertsFromPEM([]byte(certData.CACertData))
+
+		if ok := pool.AppendCertsFromPEM(decodedCACertData); !ok {
+			log.Infof("unable to use custom cert when taking to %s endpoint", c.tanzuHubEndpoint)
```
vuil, line 180: same

prkalle, line 180: same here `taking` ->`interacting`?

vuil (approved):
Other than a few typos, lgtm, thanks!

client/hub/client.go
```diff
@@ -162,12 +164,22 @@ func (c *hubClient) getTLSConfig() *tls.Config {
 	// If CACertData is present use it
 	if certData.CACertData != "" {
 		var pool *x509.CertPool
-		var err error
+
+		decodedCACertData, err := base64.StdEncoding.DecodeString(certData.CACertData)
+		if err != nil {
+			log.Infof("unable to use custom cert when taking to '%s' endpoint. Error: %s", c.tanzuHubEndpoint, err.Error())
+			return nil
```
prkalle, line 171: should it be `return err`?

anujc25: In this case, I was thinking about returning `nil` for `*tls.Config` with the log message, so interacting with the client will throw error if the cert verification fails instead of returning error here.

prkalle: Ok. Got. Thank you!

PR #218, opened 2024-11-05
[Release-1.4] CherryPick: Have the Hub UT framework send what it received

anujc25:
Cherry-pick: #206

Before this commit the UT framework would send the registered operation. This was not useful for unit tests as it is those same tests that registered the operation, so they are already aware of what it is.  What the tests need is for the framework to send the actual received GraphQL request so that it can be verified. This commit does this.

Fixes #

### Release note

```release-note
Hub unit test framework now sends back the GraphQL request it received for more thorough testing.  This change breaks backwards-compatibility of the API used to write unit test for GraphQL: the `Responder` and `EventGenerator` types have a slightly different signature.
```

### DIFF
--- a/client/hub/client_test.go
+++ b/client/hub/client_test.go
@@ -31,12 +31,21 @@ import (
 //     }
 //   }
 
+// A set of variables that are sent with the query.
+// This allows us to test that the mock server returns these
+// variables in the response.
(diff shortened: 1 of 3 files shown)


marckhouzam (approved):
LGTM
Thanks!

PR #219, opened 2024-11-21
Increases the graphql subscription client buffer size to 1mb to handle larger event chunks

mdzhigarov:

Fixes #

### Release note

```release-note
Increases the GraphQL subscription client buffer size to 1MB to handle larger event chunks
```


anujc25 (approved):
LGTM. Thanks for the fix.

PR #224, opened 2025-01-21
Add additional generic components and helpers

anujc25:
Add additional generic components and helpers functions

1. Adds Cobra command's `Arg`, `Error` and `Extend` helpers under `command` package
2. Adds Emoji and special characters under `component/icons` package
3. Adds string and color utils  under `component/stringutils` package
4. Adds TabWriter component under `component/tabwriter` package

Fixes #

### Release note

```release-note
Add Cobra command's `Arg`, `Error` and `Extend` helpers, Icons, Color printer and TabWriter component
```


command/extend.go
```diff
@@ -0,0 +1,48 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+type commandKey struct{}
+
+func WithCommand(ctx context.Context, cmd *cobra.Command) context.Context {
```
anujc25, line 39: Should we rename this to `WithContext`? So, invocation will be `command.WithContext(ctx, cmd)` ?

command/extend.go
```diff
@@ -0,0 +1,48 @@
...
+	return context.WithValue(ctx, commandKey{}, cmd)
+}
+
+func CommandFromContext(ctx context.Context) *cobra.Command {
```
anujc25, line 43: Should we rename this to `FromContext`? So, invocation will be `command.FromContext(ctx)` ?

command/arguments.go
```diff
@@ -0,0 +1,152 @@
...
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+const (
+	NameArgumentName  = "name"
+	NamesArgumentName = "name(s)"
+)
```
anujc25, line 15: Are these generic enough to be kept here? Also applies to NameArg, OptionalNameArg, NamesArg functions below.

component/printer/color.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+var (
+	FaintColor   = color.New(color.Faint)
+	InfoColor    = color.New(color.FgCyan)
+	SuccessColor = color.New(color.FgGreen)
+	WarnColor    = color.New(color.FgYellow)
+	ErrorColor   = color.New(color.FgRed)
+	BoldColor    = color.New(color.Bold)
+)
```
anujc25, line 20: Noticed that we already have `compnent/colorable-tty.go` which holds similar color specific helpers. Check to see how we can combine both.

component/characters/icon.go
anujc25, line 1: Any suggestion on the package name `characters`?

component/printer/color.go
anujc25: I have deprecated some of the string helpers in `colorable-tty.go` and created a new package `stringutils` which includes all the string specific utils functions along with `color` specific helpers.

command/extend.go
anujc25: I have updated the function accordingly but keep this comment open if others have any opinions.

command/arguments.go
marckhouzam, line 1: There is a slight update to this file in the apps plugin.
The idea of that change was to allow the caller to specify the string to use for the argument name, instead of forcing it to `name`.
Also, but it is a matter of opinion, the above change renames some of the function to something that I felt was clearer to understand.

command/extend.go
marckhouzam: I agree with you.  Calling `command.WithCommand()` does not indicated at all that we are getting a context.  `command.ContextWithCommand()` is better.

command/arguments.go
marckhouzam: Please see the comment about that file.

I would keep those two constants to let other call them if they want to use a standard format.

marckhouzam (commented):
Nice stuff.
Thanks for the effort!

command/arguments.go
anujc25: Yes. I have tried to add those as well. See additional functions I have added here: 
```
func Argument(name string, val *string) Arg {
func OptionalArgument(name string, val *string) Arg {
func RemainingArguments(name string, values *[]string) Arg {
func OptionalRemainingArguments(name string, values *[]string) Arg {
```

Are there any more that you think I missed?

component/stringutils/color.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/fatih/color"
+
+	"github.com/vmware-tanzu/tanzu-plugin-runtime/component"
+)
+
+func init() {
+	// Configure the global `NoColor` option within github.com/fatih/color library
+	// based on the user's terminal and provided options
+	color.NoColor = !component.IsTTYEnabled()
```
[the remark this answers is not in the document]
anujc25: We have additional `TANZU_CLI_NO_COLOR` command that we are supporting with `IsTTYEnabled` and hence I thought I would just use it.

We can also include the same Cygwin check in `IsTTYEnabled` as well if that makes sense.

component/characters/spaces.go
```diff
@@ -0,0 +1,12 @@
+// Copyright 2025 VMware, Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package characters provides common set of characters for spaces and icons
+package characters
+
+const (
```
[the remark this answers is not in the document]
anujc25: I will remove this from here.

component/characters/icon.go
anujc25: I am going to rename this to `icons` as it only contains icons (emojis) now.

component/stringutils/color.go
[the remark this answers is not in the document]
marckhouzam: I missed the `TANZU_CLI_NO_COLOR`.  Thanks for noticing.
What we could do is
	`color.NoColor = color.NoColor || !component.IsTTYEnabled()`
to cover both cases.

P.S. I assume that the creation of the variable AND setting it in fatih/color both happen before this `init()` call, but I'm not an expert at Go init ordering.

command/arguments.go
marckhouzam: Oh sorry, I missed that you kept both series.
When I made the change `Argument(..)` was meant to replace `NameArg(...)` to allow to set the name of the arg and because I found the term `NameArg` unclear, especially with the argument to the function being `name`, which is totally confusing to me.

If you like keeping the original functions also, which I do see the value in, to avoid specifying the `NameArgumentName` argument in every call, I would suggest a couple of things:
1- change the arguments from `name` and `names` to `val` and `values`, which I think are clearer (and would match the other functions)
2- have the more specific functions call the more generic ones, e.g.:
```
func NameArg(name *string) Arg {
	return Argument(NameArgumentName, val)
}
```

component/README.md
```diff
@@ -36,26 +36,30 @@ The function takes a message string as its argument and returns an error if the
 
 If an error occurs during the function's execution, it will return a new error wrapped with the `github.com/pkg/errors` package.
 
-## Colorable-TTY
+## StringUtils (includes string colors and formatting)
 
-`Colorable-TTY` is a Golang package that provides utilities for creating colorized and formatted console output. This package uses the `logrusorgru/aurora` package to generate the colorized output. It also includes several functions for formatting strings such as `Rpad`, `Underline`, `Bold`, `TrimRightSpace`, and `BeginsWith`.
+This package that provides string utility functions for creating colorized and formatted console output. This include following predefined formatting options like `Faint`, `Bold` , `Underline` and `Italic` as well as color options like `InfoColor`, `SuccessColor`, `WarnColor` and `ErrorColor`.
```
marckhouzam, line 41: I suggest small changes to this sentence:

The `stringutils` package provides string utility functions for creating colorized and formatted console output. These include the following predefined formatting options like `Faint`, `Bold` , `Underline` and `Italic` as well as color options like `InfoColor`, `SuccessColor`, `WarnColor` and `ErrorColor`.

component/terminal.go
```diff
@@ -0,0 +1,29 @@
...
+	"io"
+	"strings"
+)
+
+// From https://github.com/kubernetes/cli-runtime/blob/v0.28.1/pkg/printers/terminal.go"
+
+// terminalEscaper replaces ANSI escape sequences and other terminal special
+// characters to avoid terminal escape character attacks (https://github.com/kubernetes/kubernetes/issues/101695).
+// Add "\x1b", "^[" to the `NewReplacer` params to escape color
+var terminalEscaper = strings.NewReplacer("\x1b", "^[", "\r", "\\r")
```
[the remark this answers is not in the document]
anujc25: Yeah. I was thinking as well and was going to ask you on whether we should add or not. But I would remove it for now to be on safer-side.

marckhouzam (approved):
LGTM
Very nice.
If there are things to tweak when we start using this change in an existing plugin, we can fix in another PR.

PR #227, opened 2025-02-18
Upgrade to Cobra 1.9.1

marckhouzam:
### What this PR does / why we need it

Upgrade to Cobra 1.9.1

### Which issue(s) this PR fixes

Fixes # N/A

### Describe testing done for PR

Let CI run

### Release note

```release-note
Upgrade to Cobra 1.9.1
```

PR #228, opened 2025-02-18
Don't use templating to allow for dead-code elimination

marckhouzam:
This PR is on top of #227 so that the new test that checks for dead-code elimination works (or else the old Cobra version would prevent dead-code elimination and the test would fail).

When templating is used, the linker cannot know which functions are called and which one are not; this is because templating use calls to `MethodByName()` which can be used to call any function (similar to reflection). 

With the release of Cobra 1.9.1, templates are no longer used by default in Cobra. By also not using templates in the tanzu-plugin-runtime it now gives an opportunity for plugins to try to avoid templates to allow dead-code elimination to work, if they so choose.

This PR therefore removes the use of templates in tanzu-plugin-runtime.
Note that we had technically already moved to using go code but we just had kept the template engine calling the go code.

Note that the new `TestDeadcodeElimination` test is meant to notify us if templates are added.  However, if any new dependency used by tanzu-plugin-runtime uses templates or reflection, it would also turn off dead-code elimination and fail the test.  If there is a justification for that to happen, we will just need to disable the test.

Ref: https://github.com/spf13/cobra/pull/1956

Fixes # N/A

I built a plugin using this new PR and verified that both its usage and help printouts were not changed:
```
$ tz project list -h
List the projects in the current workspace.

Project list configuration options include:
- Output the project formatted

Usage:
  tanzu project list [flags]

Aliases:
  list, ls

Examples:
  tanzu project list
  tanzu project list --wide

Flags:
  -h, --help   help for list
  -w, --wide   output the project list with additional information
$ tz project list --invalid
Usage:
  tanzu project list [flags]

Aliases:
  list, ls

Examples:
  tanzu project list
  tanzu project list --wide

Flags:
  -h, --help   help for list
  -w, --wide   output the project list with additional information

Error: unknown flag: --invalid
```

### Release note

```release-note
Do not use templates so that dead-code elimination is not turned off.
```

### DIFF
--- a/plugin/root.go
+++ b/plugin/root.go
@@ -45,8 +45,12 @@ func newRootCmd(descriptor *PluginDescriptor) *cobra.Command {
 			cobra.CommandDisplayNameAnnotation: cmdName,
 		},
 	}
+	// Instead of using templates, use go functions.
+	// This allows for dead-code-elimination.
+	// The below call will set the format for both usage and help printouts.
+	cmd.SetUsageFunc(UsageFunc)
+	// Keep this call for backwards-compatibility, in case a plugin uses the templating
 	cobra.AddTemplateFuncs(TemplateFuncs)
-	cmd.SetUsageTemplate(cmdTemplate)
 
 	cmd.AddCommand(
 		newDescribeCmd(descriptor.Description),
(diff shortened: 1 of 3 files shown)


prkalle (approved):
Nice changes!!

LGTM. Thanks!

